PR #35, opened 2023-06-22
Add a Transformer to Expand Embedded JSON fields

### LINKED ISSUE #2: [REQUEST] JSON Field Expansion
Far too prevalent in data coming from a Data Warehouse or even APIs, you can find data that is a JSON string encoded into a String field in a column or JSON response. There is no inbuilt way to "inflate" these values to a workable value.  

I think building in a basic transformer to handle these cases would be ideal. Ideally the interface for the transformer would be something like this:

```yaml
- implementation: nodestream.pipeline:ExpandJsonFields
  arguments:
     path: key_of_embedded_json
```

or this:

```yaml
- implementation: nodestream.pipeline:ExpandJsonFields
  arguments:
     path: [ "deep", "key", "to", "embedded", "json"] 
```

Whereby the embedded string is simply replaced with the inflated JSON as a dictionary, list, or other. 

Currently, the only alternative is to do data preprocessing or write this transformer yourself. 

**Additional context**
n/a

zprobst:
closes #2


zprobst:
this requires an update to the reference

PR #44, opened 2023-07-09
Overhauling Pluggability

### LINKED ISSUE #10: [REQUEST] Implement Pipeline Sharing and Distribution Solution

Part of `nodestream`s promise is that it allows us a way to build an ecosystem of ingests that can be used by many people. To accomplish this, we need a mechanism by which people can (including nodestream-proj) itself to distribute complete pipelines to users. An ideal solution would be for `nodestream` to have some machinery such that you can `pip install nodestream-my-pipeline-package` and get new pipelines importable into your project. 

I think the simplest way to success on this is hook into the `Project` import machinery. Consider (`pytest`)[https://docs.pytest.org/en/7.1.x/how-to/writing_plugins.html#registering-custom-markers] as an example. `pytest` looks for a `pytest_configure` in modules that it imports and passes a configuration object. I think a similar solution work for us as well. Technically, this could look like like this:

`Project` has a method called `ensure_modules_are_imported`. This method could be extended to import and call a hook named `nodestream_configure(project: Project)` if it has one.  From there, plugins can use [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html) to add scopes of pipelines to the users project automatically. 

While its easy enough to distribute extensions to nodestream (`ValueProvider`s, `Normalizer`s, etc.), python does not provide an obvious way to handle distributing pipelines 

**Additional context**

This leaves out one glaring issue. How does the pipeline know what database you are using provide a sensible configuration? A pipeline includes the definition of [the database configuration](https://github.com/nodestream-proj/nodestream/blob/main/playground_project/test2.yaml#L14) which by design is part of the pipeline.

zprobst:
`nodestream` previously had a plugins mechanism called `imports` that was housed in the project file. This PR does a few things:

- Implements a `ProjectPlugin` mechanic that allows pluggable additions or modifications to one's project. 
- Documents the `Project` apis better so it is easier to implement `ProjectPlugin`s.
- Replaces the `imports` section in the `nodestream.yaml` file with a system that uses Python `entry_points`. 
- Replaces the new command with a cookiecutter template so that it could implement it better.

This closes #10


docs/docs/guides/creating-your-own-commands.md
```diff
@@ -0,0 +1,118 @@
...
+    async def handle_async(self) -> None:
+        name = self.argument("name")
+        message = f"Hello {name}!"
+        if self.option("shout"):
+            message = message.upper()
+        self.line(message)
+```
+
+
+### Registering the Audit
```
ccloes, line 76: Should this say "Registering the Command"?

docs/docs/guides/creating-your-own-interpretation.md
```diff
@@ -61,13 +61,53 @@ class MemoizeNegativeProperty(Interpretation, alias="memoize_negative"):
...
-Wherever you have your class defined, nodestream needs to know that its something that should be imported. To do
-so, add your module to the imports section of your `nodestream.yaml` file. For example:
+
+Interpretations are registered via the [entry_points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) API of a Python Package. Specifically, the `entry_point` named `interpretations` inside of the `nodestream.plugins` group is loaded. Every Value Provider is expected to be a subclass of `nodestream.interpreting:Interpretation` as directed above. 
+
+The `entry_point` should be a module that contains at least one Value Provider class. At runtime, the module will be loaded and all classes that inherit from `nodestream.interpreting:Interpretation` will be registered. The `alias` attribute of the class will be used as as the name of the tag used in the yaml pipeline.
+
+Depending on how you are building your package, you can register your Value Provider plugin in one of the following ways:
+
+=== "pyproject.toml"
```
ccloes, line 73: These example sections are redundant, should it be abstracted out and referred to in one place?  

Maybe have a "See examples" file and link to it?

docs/docs/guides/creating-your-own-normalizer.md
```diff
@@ -18,16 +18,39 @@ class RoundToWholeNumber(Normalizer, alias="round_numbers"):
         return int(value) if isinstance(value, float) else value
 ```
 
-## Make sure your module is imported
+## Registering your Normalizer
 
-Wherever you have your class defined, nodestream needs to know that its something that should be imported. To do
-so, add your module to the imports section of your `nodestream.yaml` file. For example:
+Normalizers are registered via the [entry_points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) API of a Python Package. Specifically, the `entry_point` named `normalizers` inside of the `nodestream.plugins` group is loaded. Every Value Provider is expected to be a subclass of `nodestream.pipeline.normalizers:Normalizer` as directed above.
```
ccloes, line 23: Is Value Provider overloaded here in these instructions, or am I am not following?  Probably the later...

docs/docs/guides/customizing-the-stream-extractor.md
```diff
@@ -20,6 +29,40 @@ class FileConnector(StreamConnector, alias="file"):
 
 The `alias` class attribute is used to specify the name of the stream connector. This name is used in the `stream` configuration option of the `StreamExtractor` to specify which stream connector to use. For more information see the [Stream Extractor](../reference/extractors.md) reference.
 
+### Registering the Stream Connector
+
+Stream connectors are registered via the [entry_points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) API of a Python Package. Specifically, the `entry_point` named `stream_connectors` inside of the `nodestream.plugins` group is loaded. It is expected to be a subclass of `nodestream.pipeline.extractors.streams:StreamConnector` as directed above.
```
ccloes, line 34: Here you switched to "It is expected" rather than "Value Provider" which makes me think the previous docs contained some copy pasta.

docs/docs/guides/project-plugins.md
```diff
@@ -0,0 +1,91 @@
...
+        scope = PipelineScope.from_resources(name="my_plugin", package="my_plugin.pipelines")
+        project.add_scope(scope)
+```
+
+### Registering the Project Plugin
+
+Project plugins are registered via 
+the [entry_points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) API of a Python Package.
+Specifically, the `entry_point` named `project` inside of the `nodestream.plugins` group is loaded.
+All project plugins are expected tobe a subclass of `nodestream.project:ProjectPlugin` as directed above.
```
ccloes, line 59:
```suggestion
All project plugins are expected to be a subclass of `nodestream.project:ProjectPlugin` as directed above.
```

ccloes (approved):
Couple questions, corrections, and comments.

docs/docs/guides/creating-your-own-commands.md
zprobst: Good catch - yes!

docs/docs/guides/creating-your-own-normalizer.md
zprobst: Yeah I copy-pasta-ed here.

docs/docs/guides/creating-your-own-interpretation.md
zprobst: Yeah I agree. I will see if there is a good way to de-duplicate this that works nicely in the docs.

docs/docs/guides/customizing-the-stream-extractor.md
zprobst: Indeed.

docs/docs/guides/creating-your-own-interpretation.md
zprobst: I tried the suggestion out, but it was a little cumbersome to follow a step-by-step guide over multiple pages. Do you have a strong preference to eliminate the redundancy? I found it "familiar" that each guide follows a similar structure and gets you from start to finish.

PR #47, opened 2023-07-16
Add a Prometheus Server On Pipeline Run

### LINKED ISSUE #39: [REQUEST] Expose Ingestion Statistics as a Prometheus Server
For perpetually running pipelines in particular, its hard to get a sense of how much work the pipeline is doing against the database. It would be ideal if the user

Expand the `nodestream run` command to configure and run a prometheus metrics server:

```bash
nodestream run <<pipeline>> --prometheus-server-addr 127.0.0.1:8080
```

We already collect metrics as part of the `PipelineContext` api accessible via `get_context()` in `nodestream.pipeline.meta`. This means that we can reuse a lot of existing logic for Prometheus. As well as pipeline or application level metrics, it would be ideal to include some process/system level metrics such as memory and cpu usage.

Currently there is max memory usage logged as part of the message but that isn't really good enough. the only other alternative is to monkey patch the system, or use custom components for everything. 

**Additional context**
n/A

zprobst:
Closes #39


ccloes (approved):
Don't see anything that stands out.

Tested and ready?

PR #49, opened 2023-07-16
SQS Connector

### LINKED ISSUE #3: [REQUEST] Create an SQS Stream Connector
SQS is a common queueing mechanism. Since this library has other AWS integrations, its reasonable that we expand it to include SQS. 

Implement a `StreamConnector` that polls AWS SQS. Ideally, it would be able to use the same assume role functionality. 

The alternative is to simply write the connector externally in another library or on your own. 

**Additional context**
N/A

zprobst:
closes #3


ccloes:
This looks like its till a work in progress.  Let me know if you are ready for a review.

PR #52, opened 2023-08-06
Refactor: Introduce Database Connectors Concept

zprobst:
Due to our future plans of having migrations and other deeper integrations with internals of the database, this PR introduces a `DatabaseConnector` class that represents the integration point between a database and nodestream. With features like the migrations and more advanced indexing support, we are likely going to need to grow the interface beyond what the `QueryExecutor` is designed to do. The new `DatabaseConnector` class can act as the top level interface a database implementation needs to implement.

In short, instead of the `GraphDatabaseWriter` creating a `QueryExecutor` directly, it creates a `DatabaseConnector` and asks the `DatabaseConnector` to create a `QueryExecutor`.

### DIFF
--- a/nodestream/databases/database_connector.py
+++ b/nodestream/databases/database_connector.py
@@ -1,0 +1,35 @@
+from abc import ABC, abstractmethod
+
+from ..pluggable import Pluggable
+from ..subclass_registry import SubclassRegistry
+from .query_executor import QueryExecutor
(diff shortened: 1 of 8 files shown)


ccloes (approved):
I started making a comment/question, then read properly.  Neo4jIndexQueryBuilder() and Neo4jIngestQueryBuilder() were close enough in name that it confused me.  I get it now.  

Looks good.

PR #55, opened 2023-08-07
Add `RegexValueProvider`

zprobst:
Adds a Value provider that works be regex-ing another value provider. The docs provide more details on usage.

### DIFF
--- a/docs/docs/reference/value-providers.md
+++ b/docs/docs/reference/value-providers.md
@@ -120,3 +120,38 @@ The following interpretation would create a node with the key `Hello, Joe!`:
         properties:
           age: !jmespath age
 ```
+
+## `!regex`
+
+The `!regex` value provider allows you to extract a value from a string using a regular expression. For example, if you wanted to extract
+the first name from a string given a record like this:
+
+```json
+{
+    "name": "Joe Smith",
+    "age": 25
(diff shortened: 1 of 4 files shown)


docs/docs/reference/value-providers.md
```diff
@@ -120,3 +120,38 @@ The following interpretation would create a node with the key `Hello, Joe!`:
...
+
+```yaml
+- implementation: nodestream.interpreting:Interpreter
+  arguments:
+    interpretations:
+      - type: source_node
+        node_type: HelloNode
+        key:
+          first_name: !regex 
+            regex: "^(?P<first_name>[a-zA-Z]+)\s(?P<last_name>[a-zA-Z]+)$"
```
ccloes, line 146: I don't understand how `<last_name>` is used here... can you help me understand that? Seems like it would simply be dropped?

ccloes (approved):
I have one question, but I assume its me misunderstanding.

docs/docs/reference/value-providers.md
zprobst: Yeah you are right. This is just indicating that you can name groups and not use them all if that makes sense. Its only the group that you set in group that gets used.

PR #59, opened 2023-08-08
Graph schema extraction

AGhafaryy:


nodestream/schema/printers/graph_schema_extraction.py
```diff
@@ -0,0 +1,25 @@
+from .graphql_schema_printer import GraphQLSchemaPrinter
+from collections import defaultdict
+
+class Neo4jLLMGraphSchemaExtraction(GraphQLSchemaPrinter):
```
zprobst, line 4: This class should extend from the `SchemaPrinter` base class and specify an alias. See the defiinition of `GraphQLSchemaPrinter`.

tests/unit/schema/printers/test_extract_schema.py
```diff
@@ -0,0 +1,99 @@
+from collections import defaultdict
```
zprobst, line 1: Can you follow the pattern for the other schema printers? 
https://github.com/nodestream-proj/nodestream/blob/77579a2fe237890c7152776ce50674cba9d5b2e7/tests/unit/schema/printers/test_graphql_schema_printer.py

nodestream/schema/printers/graph_schema_extraction.py
```diff
@@ -0,0 +1,25 @@
+from .graphql_schema_printer import GraphQLSchemaPrinter
```
zprobst, line 1: Can you add the `Neo4jLLMGraphSchemaExtraction` to the package level `__init__.py` so that its loaded and available?

zprobst (changes requested):

PR #60, opened 2023-08-08
Add `Project#get_snapshot_for` API

zprobst:
Prior to this project being open sourced, we had/have a mechanism to snapshot the results of a pipeline's execution, and compare them against known good. This brings that support into `nodestream`.


ccloes (approved):
Looks good.  Didn't see anything that stood out.

PR #89, opened 2023-09-08
Fix incorrect filtering logic

zprobst:
The filter logic was not awaiting the wrong method. Thus everything was not working correctly in the filters. This creates a test for the core logic of a filter and fixes the noted issue.


ccloes (approved):
Looks good... could we also implement a test which tests the "NotImplementedError" part of the logic where the Class is implemented incorrectly?  Not sure the value, but just a thought.

zprobst:
> Looks good... could we also implement a test which tests the "NotImplementedError" part of the logic where the Class is implemented incorrectly? Not sure the value, but just a thought.

That's not a bad idea. I'll create a follow up PR.

PR #91, opened 2023-09-13
Yield files in order in `FileExtractor`

zprobst:
In the case where you have a series of files, each file should be loaded and yielded in a predictable order so that an appropriate snapshot is able to be created. If objects are continuously yielded in various orders, it means that one can not rely on the ordering. While, for most extractors, this isn't a big deal as you are not likely to snapshot them or the extractor has an obvious ordering. However in the case of the file extractor which takes a series of globs, its not obvious the ordering. This makes it so that is the case.


ccloes (approved):
That was easy... :D

PR #122, opened 2023-10-31
Move stream poll values into stream declaration

grantleehoffman:
- The timeout and polling values are now provided to stream connectors instead of setting defaults for all extractors. This allows the ability to have multiple types of StreamConnectors with different polling and timeout settings.


ccloes (approved):
Looks good to me

PR #129, opened 2023-11-06
add logging for extractor poll message exceptions

grantleehoffman:
Add error logging for extractor poll message exceptions

### DIFF
--- a/nodestream/pipeline/extractors/streams/extractor.py
+++ b/nodestream/pipeline/extractors/streams/extractor.py
@@ -85,7 +85,7 @@ class StreamExtractor(Extractor):
             else:
                 for record in results:
                     yield self.record_format.parse(record)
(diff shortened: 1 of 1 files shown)


nodestream/pipeline/extractors/streams/extractor.py
```diff
@@ -83,5 +85,7 @@ async def extract_records(self):
             else:
                 for record in results:
                     yield self.record_format.parse(record)
+        except Exception as e:
+            self.logger.error(f"failed extracting records, Exception: '{e}'")
```
zprobst, line 89: We should probably use `self.logger.exception` here. If we just format in the error, then the we will miss some context most likely.

grantleehoffman: ok

PR #136, opened 2023-11-08
add options to disable apoc_iterate

grantleehoffman:
- Add option to disable apoc_iterate in neo4j connector

### DIFF
--- a/nodestream/databases/neo4j/database_connector.py
+++ b/nodestream/databases/neo4j/database_connector.py
@@ -19,19 +19,18 @@ class Neo4jDatabaseConnector(DatabaseConnector, alias="neo4j"):
         password: str,
         database_name: str = "neo4j",
         use_enterprise_features: bool = False,
-        apoc_iterate: bool = True,
+        use_apoc: bool = True,
     ):
         driver = AsyncGraphDatabase.driver(uri, auth=(username, password))
         if use_enterprise_features:
             index_query_builder = Neo4jEnterpriseIndexQueryBuilder()
         else:
-            index_query_builder = Neo4jIndexQueryBuilder()
+            index_query_builder = Neo4jIndexQueryBuilder(use_apoc)
         return cls(
             driver=driver,
             index_query_builder=index_query_builder,
-            ingest_query_builder=Neo4jIngestQueryBuilder(),
+            ingest_query_builder=Neo4jIngestQueryBuilder(use_apoc),
             database_name=database_name,
-            apoc_iterate=apoc_iterate,
         )
 
     def __init__(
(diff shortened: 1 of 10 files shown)


nodestream/databases/neo4j/query.py
```diff
@@ -12,41 +12,51 @@
...
 
+NON_APOCH_COMMIT_QUERY = """
+UNWIND $param_sets AS param
+CALL apoc.cypher.doIt($batched_query, {params: param}) 
+YIELD value 
+RETURN value
+"""
 
 @dataclass(slots=True, frozen=True)
 class Query:
```
zprobst, line 23: `Query` represents a single query to be submitted to the database. I don't think we need to include `apoc_iterate` rules in here.

nodestream/databases/neo4j/query.py
```diff
@@ -12,41 +12,51 @@
...
             },
+            self.apoc_iterate
         )
 
 
 @dataclass(slots=True, frozen=True)
 class QueryBatch:
     query_statement: str
     batched_parameter_sets: List[Dict[str, Any]]
+    apoc_iterate: bool
```
zprobst, line 49: We could keep `apoc_iterate` stored here. But it means that we can't extend this. Since we already pass `apoc_iterate` to the `as_query` function, we could store the batching query instead.

nodestream/databases/neo4j/index_query_builder.py
```diff
@@ -10,28 +10,28 @@
...
     )
-    return Query.from_statement(statement)
+    return Query.from_statement(statement, apoc_iterate)
 
 
 class Neo4jIndexQueryBuilder:
     """Creates index creation queries that will work for any Neo4j Database Supported."""
 
-    def create_key_index_query(self, key_index: KeyIndex) -> Query:
+    def create_key_index_query(self, key_index: KeyIndex, apoc_iterate: bool) -> Query:
```
zprobst, line 27: I don't think we should have to change the `index_query_builder` module since it doesn't do anything with batching.

nodestream/databases/neo4j/database_connector.py
```diff
@@ -19,6 +19,7 @@ def from_file_data(
         password: str,
         database_name: str = "neo4j",
         use_enterprise_features: bool = False,
+        apoc_iterate: bool = True,
```
zprobst, line 22: I wonder if at the connector level we should name this more generically. Like `use_apoc`. It could be a handy flag for other uses of the `apoc` library in the future. What do you think? Too generic?

nodestream/databases/neo4j/extractor.py
```diff
@@ -18,6 +18,7 @@ def __init__(
         self.connector = Neo4jDatabaseConnector.from_file_data(
             **database_connector_args
         )
+        self.apoc_iterate = self.connector.use_apoc
```
zprobst, line 21: Do we need this? Appears unused.

zprobst (approved):
Minor comments

zprobst (approved):

PR #137, opened 2023-11-09
Augment Interpreter and Interpretation Docs with Examples and Visuals

zprobst:
Improved documentation by correcting inaccuracies and providing better discussion and examples.


docs/docs/reference/interpretations.md
```diff
@@ -2,6 +2,59 @@
...
+
+```yaml
+- type: source_node
+  node_type: Person
+  key:
+    name: John
+```
+
+Nodestream will automatically create indexes and/or constraints for the node key fields.
+As mentioned, you can also uses value providers to define the key.
```
grantleehoffman, line 26:
```suggestion
As mentioned, you can also use value providers to define the key. 
```

docs/docs/reference/interpretations.md
```diff
@@ -2,6 +2,59 @@
...
+Nodestream will automatically create indexes and/or constraints for the node key fields.
+As mentioned, you can also uses value providers to define the key. 
+This is most often what you will do. 
+As an example, if we wanted to define a source node of type `Person` with a `name` field where the value of the `name` key is defined by a field called `p_name` in your input JSON, we would use the following interpretation:
+
+```yaml
+- type: source_node
+  node_type: Person
+  key:
+    name: !jmespath patient_name
```
grantleehoffman, line 34:
```suggestion
    name: !jmespath p_name
```

grantleehoffman (commented):
This is awesome! Added in comments for a few typo's I found.

docs/docs/reference/interpretations.md
```diff
@@ -13,6 +66,125 @@
...
+```
+
+This would allow us to have multiple `PLAYED_IN` relationships between the same source node and related node as long as they have different `role` properties.
+
+### Relationship Match Strategy
+
+By default, related nodes are created when not present based on the supplied node type and key. However, you can change this behavior by setting the `match_strategy` field. The `match_strategy` field is a string that can be one of the following values:
+
+* `EAGER` - Related nodes will be created when not present based on the supplied node type and key. (Default)
+* `MATCH_ONLY` - Will not create a relationship when the related node does not already exists.
```
ccloes, line 150:
```suggestion
* `MATCH_ONLY` - Will not create a relationship when the related node does not already exist.
```

docs/docs/reference/interpretations.md
```diff
@@ -13,6 +66,125 @@
...
+* `FUZZY` - Behaves like `MATCH ONLY`, but treats the key values as regular expressions to match on.
+
+### Working with Many Relationships
+
+By default, the cardinality of a relationship is one-to-one (or many; depending on the data). 
+This means that the relationship interpreation will only relate to one node per instance of the interpreation.
```
ccloes, line 156:
```suggestion
This means that the relationship interpretation will only relate to one node per instance of the interpretation. 
```

docs/docs/reference/interpretations.md
```diff
@@ -13,6 +66,125 @@
...
+
+* `EAGER` - Related nodes will be created when not present based on the supplied node type and key. (Default)
+* `MATCH_ONLY` - Will not create a relationship when the related node does not already exists.
+* `FUZZY` - Behaves like `MATCH ONLY`, but treats the key values as regular expressions to match on.
+
+### Working with Many Relationships
+
+By default, the cardinality of a relationship is one-to-one (or many; depending on the data). 
+This means that the relationship interpreation will only relate to one node per instance of the interpreation. 
+Often times when working with document data, you might encounter a list of values all of which should be related to the source node.
```
ccloes, line 157:
```suggestion
Oftentimes when working with document data, you might encounter a list of values all of which should be related to the source node. 
```

docs/docs/reference/interpretations.md
```diff
@@ -13,6 +66,125 @@
...
+```
+
+
 | Parameter Name          	| Required? 	| Type                               	| Description                                                                                                                                                                                                                                                                                                            	|
 |-------------------------	|-----------	|------------------------------------	|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------	|
 | node_type               	| Yes       	| String or ValueProvider            	| Specifies the type of the node to relationship connects to. It is a required field. When a ValueProvider is used dynamic index creation and schema introspection are not supported.                                                                                                                                    	|
```
ccloes, line 190:
```suggestion
| node_type               	| Yes       	| String or ValueProvider            	| Specifies the type of the node a relationship connects to. It is a required field. When a ValueProvider is used dynamic index creation and schema introspection are not supported.                                                                                                                                    	|
```

docs/docs/reference/interpretations.md
```diff
@@ -47,6 +249,25 @@
 
 The switch interpretation allows you to define multiple interpretations that will be applied. The interpretations that will be applied is determined by the value of the `switch_on` parameter. The value of the `switch_on` parameter is a value provider that will be evaluated for each source node. The value of the value provider will be used to determine which interpretation to apply. The interpretation that will be applied is the one that has the same value as the value of the `switch_on` parameter. If no interpretation has the same value as the value of the `switch_on` parameter, the default interpretation will be applied. If no default interpretation is defined, the source node will be ignored.
```
ccloes, line 250:
```suggestion
The switch interpretation allows you to define multiple interpretations that will be applied. The interpretation that will be applied is determined by the value of the `switch_on` parameter. The value of the `switch_on` parameter is a value provider that will be evaluated for each source node. The value of the value provider will be used to determine which interpretation to apply. The interpretation that will be applied is the one that has the same value as the value of the `switch_on` parameter. If no interpretation has the same value as the value of the `switch_on` parameter, the default interpretation will be applied. If no default interpretation is defined, the source node will be ignored.
```

docs/docs/reference/interpreter.md
```diff
@@ -52,63 +71,87 @@ Sometimes single records come in a format were multiple ingestions are required
...
+            address: !jmespath address
 ```
 
 Note that in the `interpretations` section we are using `!jmespath` that are relative to the current object in the iteration. In the `before_iteration` section we are using `!jmespath` that are relative to the source object.
 
 ## Performing Multiple Passes
 
-One limitation is that the interpreter can only relate nodes one level deep. For example, if we want to relate a person to a team, and a team to an address,  and an address to a city, we can't do that in a single pass. To accomplish it, we can do this by performing multiple passes over the data. In the following example, we will perform two passes. The first pass will relate a person to a team, and the second pass will relate a team to an address.
+One limitation is that the interpreter can only relate nodes one level deep. For example, if we want to relate a person to an address, and a address to a city, we can't do that in a single pass. To accomplish it, we can do this by performing multiple passes over the data. In the following example, we will perform two passes. The first pass will relate a person to a team, and the second pass will relate a team to an address.
```
ccloes, line 112:
```suggestion
One limitation is that the interpreter can only relate nodes one level deep. For example, if we want to relate a person to an address, and an address to a city, we can't do that in a single pass. To accomplish it, we can do this by performing multiple passes over the data. In the following example, we will perform two passes. The first pass will relate a person to a team, and the second pass will relate a team to an address.
```

ccloes (dismissed):
Some typos

PR #163, opened 2023-11-18
Add regex filter

grantleehoffman:
Add the ability to filter a record based on a regex (include/exclude)

### DIFF
--- a/tests/unit/pipeline/test_filters.py
+++ b/tests/unit/pipeline/test_filters.py
@@ -4,7 +4,7 @@ from hamcrest import assert_that, equal_to
 from nodestream.pipeline.filters import (
     ExcludeWhenValuesMatchPossibilities,
     Filter,
-    ValueMatchesRegex,
+    ValueMatchesRegexFilter,
     ValuesMatchPossibilitiesFilter,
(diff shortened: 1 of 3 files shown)


tests/unit/pipeline/test_filters.py
```diff
@@ -77,3 +78,68 @@ async def filter_record(self, record):
...
+    assert_that(result, equal_to(True))
+
+
+@pytest.mark.asyncio
+async def test_exclude_match_regex_failing():
+    subject = ValueMatchesRegexFilter.from_file_data(
+        fields=FAILED_MATCH_EXCLUDE_REGEX_CONFIGURATION
+    )
+    result = await subject.filter_record({})
+    assert_that(result, equal_to(False))
```
zprobst, line 145: If we want to DRY this test up, we could [parametrize it](https://github.com/nodestream-proj/nodestream/blob/355cdf677e81f80fa07280060f085d5afdfa1a97/tests/unit/pipeline/normalizers/test_trim_whitespace.py#L7) Not required, just a suggestion

zprobst (approved):

PR #176, opened 2023-12-11
Adding a single timeout extension

angelosantos4:
When the neo4j writer is initialized, it verifies a connection to the database. In this process it yields the processor to the event loop moving it to the extractor/transformers. If these upstream steps take too long to yield the processor back to the neo4j client, the socket will time out throwing an exception. This is a protection for that edge case and the exception will be thrown again if the connection genuinely has a ServiceUnavailable exception. 

Added a fix for the concurrent transformer's greediness on the processor. After records are yielded, the processor is given to the other tasks on the event loop.


nodestream/databases/neo4j/query_executor.py
```diff
@@ -81,7 +82,16 @@ async def execute(self, query: Query, log_result: bool = False):
                 "uri": self.driver._pool.address.host,
             },
         )
-        await self.driver.verify_connectivity()
+
+        try:
+            await self.driver.verify_connectivity()
```
yasonk, line 87: The explanation in the description of the ticket could go in here as a long term comment

nodestream/pipeline/transformers/transformer.py
```diff
@@ -84,13 +84,16 @@ def drain_completed_tasks():
                 # Flush the pending tasks, then yield the flush.
                 # In order to fully respect the Flush,
                 # we need to wait for all pending tasks to complete.
+                # If we're still waiting we can yield the processor.
                 while pending_tasks:
                     for result in drain_completed_tasks():
                         yield result
+                    self.yield_processor()
```
yasonk, line 91: I wonder if the same result would be achieved by adding `await`:
```
for result in await drain_completed_tasks_async(pending_tasks):
```
I believe calling with an `await` will pass control back to the event loop.

nodestream/pipeline/transformers/transformer.py
```diff
@@ -107,6 +111,9 @@ def drain_completed_tasks():
             for result in drain_completed_tasks():
```
yasonk, line 111: This looks like another contender for blocking the event loop. Should we make this also async friendly?

nodestream/pipeline/transformers/transformer.py
```diff
@@ -99,6 +102,7 @@ def drain_completed_tasks():
                         submitted = True
                     for result in drain_completed_tasks():
```
yasonk, line 103: This code seems to repeat 3 times maybe something like this would work:
```
async def drain_completed_tasks_async(pending_tasks):
    for result in drain_completed_tasks():
        yield result
    await asyncio.sleep(0)
```

Then inside this function try to use await like in the comment above?
```
for result in await drain_completed_tasks_async(pending_tasks):
```

nodestream/databases/neo4j/query_executor.py
yasonk: Also I wonder what is the purpose of `verify_connectivity` call given that the next line is going to `execute_query`. Per the docs, `verify_connectivity` does the same thing as other methods that require connectivity:
https://neo4j.com/docs/api/python-driver/current/async_api.html#neo4j.AsyncDriver.verify_connectivity

yasonk: I'm guessing it is to establish connection so that the `execute_query` call below is faster...

nodestream/databases/neo4j/query_executor.py
```diff
@@ -81,7 +82,16 @@ async def execute(self, query: Query, log_result: bool = False):
...
+        except ServiceUnavailable:
```
yasonk, line 88: Another interesting find is that when there is a time out while establishing a connection it seems that neo4j is supposed to throw a ClientError
https://github.com/neo4j/neo4j-python-driver/blob/e2448ec6e044f3aea8ee81ab6df99f37e61a2e36/src/neo4j/_async/io/_pool.py#L313

nodestream/databases/neo4j/query_executor.py
angelosantos4: Yeah it also provides this race condition to get back to the driver before the session times out too causing all other parts of the program to have to yield the processor within 60 seconds. In the event that the database is overwhelmed/unavailable I believe this would be more useful wrapped around a backoff/retry module.

nodestream/databases/neo4j/query_executor.py
angelosantos4:
```
Traceback (most recent call last):
  File ".venv/lib/python3.11/site-packages/neo4j/_async_compat/network/_bolt_socket.py", line 386, in connect
    s = await cls._connect_secure(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/neo4j/_async_compat/network/_bolt_socket.py", line 244, in _connect_secure
    raise ServiceUnavailable(
neo4j.exceptions.ServiceUnavailable: Timed out trying to establish connection to ResolvedIPv6Address(('::1', 7687, 0, 0))

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "nodestream/nodestream/databases/neo4j/query_executor.py", line 87, in execute
    await self.driver.verify_connectivity()
  File ".venv/lib/python3.11/site-packages/neo4j/_async/driver.py", line 965, in verify_connectivity
    await self._get_server_info()
  File ".venv/lib/python3.11/site-packages/neo4j/_async/driver.py", line 1179, in _get_server_info
    return await session._get_server_info()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/neo4j/_async/work/session.py", line 177, in _get_server_info
    await self._connect(READ_ACCESS, liveness_check_timeout=0)
  File ".venv/lib/python3.11/site-packages/neo4j/_async/work/session.py", line 135, in _connect
    await super()._connect(
  File ".venv/lib/python3.11/site-packages/neo4j/_async/work/workspace.py", line 178, in _connect
    self._connection = await self._pool.acquire(**acquire_kwargs_)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/neo4j/_async/io/_pool.py", line 537, in acquire
    return await self._acquire(
           ^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/neo4j/_async/io/_pool.py", line 322, in _acquire
    return await connection_creator()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/neo4j/_async/io/_pool.py", line 174, in connection_creator
    connection = await self.opener(
                 ^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/neo4j/_async/io/_pool.py", line 511, in opener
    return await AsyncBolt.open(
           ^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/neo4j/_async/io/_bolt.py", line 381, in open
    await AsyncBoltSocket.connect(
  File ".venv/lib/python3.11/site-packages/neo4j/_async_compat/network/_bolt_socket.py", line 424, in connect
    raise ServiceUnavailable(
neo4j.exceptions.ServiceUnavailable: Couldn't connect to localhost:7687 (resolved to ('[::1]:7687', '127.0.0.1:7687')):
Timed out trying to establish connection to ResolvedIPv6Address(('::1', 7687, 0, 0))
Timed out trying to establish connection to ResolvedIPv4Address(('127.0.0.1', 7687))
```

This was the stacktrace provided when the timeouts were happening. Maybe deeper in the driver source it was a different type but it was eventually rethrown as a ServiceUnavailable error when it got to the driver code.

nodestream/pipeline/transformers/transformer.py
angelosantos4: Yup good catch, this is the same as the above 2 cases.

nodestream/pipeline/transformers/transformer.py
angelosantos4: I believe making the function drain_completed_tasks_async turned it into an `AsyncGenerator` object which works on 
`async for result in drain_completed_tasks_async(pending_tasks):`

nodestream/databases/neo4j/query_executor.py
yasonk: Make sense. Looks like there is a variety of undefined issues that can cause `ServiceUnavailable`.

nodestream/databases/neo4j/query_executor.py
```diff
@@ -81,7 +82,16 @@ async def execute(self, query: Query, log_result: bool = False):
...
+            self.logger.exception(
+                "Neo4j Session timed out while waiting for other steps to resolve. Trying one more time.",
```
yasonk, line 90: Sorry, final thought on this :). Since there is a variety of issues that could cause this, we should probably have the message be less assertive on the timeout:
```
Neo4j Driver failed to connect for an unspecified reason. Maybe due to waiting for event loop to pass back control. Trying one more time.
```

yasonk (dismissed):

zprobst (approved):

zprobst (approved):

PR #178, opened 2023-12-13
Redesign Project Load Behavior for Delayed Evaluation

zprobst:
This PR introduces two related changes to improve the  approach taken to load the project file:

- It allows `ProjectPlugins` to implement a `before_project_loaded` and `after_project_loaded` hooks to better time changes to the project that the plugin may wish to make. 
- Introduce a delayed loading methodology that allows tags that are used for resolvers to be loaded after the fact for targets and scopes. This allows one to introspect the project without loading all required configuration.

### DIFF
--- a/nodestream/file_io.py
+++ b/nodestream/file_io.py
@@ -163,15 +163,9 @@ class LazyLoadedTagSafeLoader(SafeLoader):
     pass
 
 
-def construct_undefined(self, node):
-    if isinstance(node, ScalarNode):
-        value = self.construct_scalar(node)
-    elif isinstance(node, SequenceNode):
-        value = self.construct_sequence(node)
-    elif isinstance(node, MappingNode):
-        value = self.construct_mapping(node)
-    print(node.tag)
+def wrap_unloaded_tag(self, node):
+    value = self.construct_scalar(node)
     return LazyLoadedArgument(node.tag[1:], value)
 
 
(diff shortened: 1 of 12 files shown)


nodestream/file_io.py
```diff
@@ -125,3 +127,37 @@ def write_to_file(self, file_path: Path):
...
+#
+# This comes with the side effect that we shift the responsibility of resolving the arguments from the
+# yaml parser to the code that uses the "loaded" yaml data.
+class LazyLoadedArgument:
+    def __init__(self, tag, value):
+        self.tag = tag
+        self.value = value
+
+    def get_value(self):
+        return ArgumentResolver.resolve_arugment_with_alias(self.tag, self.value)
```
yasonk, line 145: typo in word `argument`

yasonk: Does this need any error handling?

zprobst: Fixing the typo caused this to be hidden in the diff. My bad. 

re: error handling

Right now, if an exception is thrown (because the tag was wrong or something was thrown trying to resolve it), it would percolate all the way out to the CLI which does a pretty good job of formatting errors and showing where the failure happened. 

If anything, we can catch and raise an exception that has better context (Such as what tag and value it was).

grantleehoffman (approved):

yasonk (approved):

PR #179, opened 2023-12-15
Plugin configs and targets

grantleehoffman:
- Reconfigure plugin schema.
- Add targets to plugin and pipeline options
- Lazy load argument resolvers for plugins


nodestream/project/plugin.py
```diff
@@ -0,0 +1,40 @@
+from dataclasses import dataclass
+from typing import List
+
+from ..file_io import LoadsFromYamlFile
+from ..pipeline.scope_config import ScopeConfig
+
+
+@dataclass
+class PluginScope(LoadsFromYamlFile):
```
zprobst, line 9: At first I wasn't sure the difference between this class and a `PipelineScope`. Its obvious when looking at the code. 

However, I think what tripped me up was the name here. What about `PluginConfiguration` Given the documentation, that seems to align with the purpose of the class. 

> A `PluginScope` represents a collection of configuration for a plugin.

nodestream/project/project.py
```diff
@@ -59,9 +59,9 @@ def describe_yaml_schema(cls):
                 Optional("scopes"): {
                     str: PipelineScope.describe_yaml_schema(),
                 },
-                Optional("plugin_config"): {
-                    str: ScopeConfig.describe_yaml_schema(),
-                },
+                Optional("plugins"): [
+                    PluginScope.describe_yaml_schema(),
+                ],
```
zprobst, line 64: I definitely like this better.

nodestream/project/plugin.py
```diff
@@ -0,0 +1,40 @@
...
+    """A `PluginScope` represents a collection of configuration for a plugin.
+
+    A config is a key value pair object in a nodestream scope including plugins.
+    It contains a collection of configuration key value pairs to be used by the pipelines of a scope.
+    """
+
+    name: str
+    config: ScopeConfig = None
+    targets: List[str] = None
```
zprobst, line 18: I wonder if we should store the `targets` as the actual `Target` objects. It does provide some cumbersome api to load the targets after listing the targets. Perhaps this is something for a future PR.

zprobst (commented):
Looks good - minor considerations and questions

nodestream/cli/operations/run_pipeline.py
```diff
@@ -45,23 +55,38 @@ def get_writer_steps_for_specified_targets(self, command: NodestreamCommand):
...
+    def combine_target_lists_without_duplicates(
+        self, pipeline_targets: List[str] = [], command_targets: List[str] = []
+    ):
+        if not pipeline_targets:
+            return command_targets
+        in_first = set(pipeline_targets)
+        in_second = set(command_targets)
+        in_second_but_not_in_first = in_second - in_first
+        result = pipeline_targets + list(in_second_but_not_in_first)
+        return result
```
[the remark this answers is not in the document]
grantleehoffman: Ahh, I was looking for this function!

nodestream/pipeline/pipeline_file_loader.py
```diff
@@ -25,7 +25,7 @@ class PipelineFileSafeLoader(SafeLoader):
     was_configured = False
 
     @classmethod
-    def configure(cls, config: ScopeConfig = None):
+    def configure(cls, config: ScopeConfig):
         if config:
```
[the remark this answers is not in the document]
grantleehoffman: This was left over after some refactoring, I'll clean it up

nodestream/project/plugin.py
grantleehoffman: Yeah I was struggling for what to call this. Works for me! I'll update it.

nodestream/pipeline/argument_resolvers/include_file_resolver.py
```diff
@@ -8,4 +8,4 @@ class IncludeFileResolver(ArgumentResolver, alias="include"):
     def resolve_argument(file_path: str):
         from ..pipeline_file_loader import PipelineFileSafeLoader
 
-        return PipelineFileSafeLoader.load_file_by_path(file_path)
+        return PipelineFileSafeLoader.load_file_by_path(file_path, None)
```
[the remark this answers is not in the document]
grantleehoffman: Nope, I think this was changed in refactoring and I didn't just use the default, i'll update.

nodestream/project/plugin.py
grantleehoffman: Yeah we would have to do some refactoring I think as we would have to pass in the Project to the plugin to get target configuration from the project, but we are also setting the plugin in the project. Not sure it gives us much storing the object here instead of the identifying string.

grantleehoffman: We could load the target schema like we do for the plugins and scope config 🤔

grantleehoffman: I think we can work on this in a future refactor since this is mostly a code cleanup task and doesn't effect the functionality of the current runtime.

zprobst: Works for me

zprobst (approved):

PR #182, opened 2023-12-18
Delayed Pipeline Argument Loading

zprobst:
This PR implements two major enhancements to the way pipelines are loaded. 

1. Implements delayed evaluation of the `arguments` section of the pipeline file until the `arguments` are actually needed. 
2. In order to do `1`, it changes the pipeline file loading logic to use the `file_io` module. This means that this PR also implements a first pass at pipeline file validation by ensuring that the file itself is a set of appropriate set definitions (`arguments`, `implementation`, and `annotations`)

### DIFF
--- a/nodestream/pipeline/class_loader.py
+++ b/nodestream/pipeline/class_loader.py
@@ -1,8 +1,6 @@
 from importlib import import_module
 from typing import Optional, Type
 
-from ..file_io import LazyLoadedArgument
-
 DECLARATIVE_INIT_METHOD_NAME = "from_file_data"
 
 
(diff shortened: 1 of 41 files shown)


nodestream/pipeline/class_loader.py
```diff
@@ -57,9 +59,10 @@ def find_class_initializer(self, implementation, factory=None):
 
     def load_class(self, implementation, arguments=None, factory=None):
         arguments = arguments or {}
+        resolved_arguments = LazyLoadedArgument.resolve_if_needed(arguments)
```
zprobst, line 62: Maybe this should move back out to the file loader to keep the responsibility out of  `class_loader`. Class loader technically doesn't know anything about yaml. what do you think?

grantleehoffman: Yeah that does make more sense

angelosantos4 (approved):

Issue #148, opened 2023-11-15
[REQUEST] Provide Friendlier Errors for `MissingFromRegistryError`s and `KeyError`s


zprobst:
**Is your feature request related to a problem? Please describe.**
Its not obvious from this error message _what_ is not found and what the pipeline is trying to load that it can. 

```

  KeyError

  None

  at ~\AppData\Local\pypoetry\Cache\virtualenvs\test1-WxuTdTEg-py3.11\Lib\site-packages\nodestream\subclass_registry.py:60 in get
       56│
       57│     def get(self, name):
       58│         """Get a subclass by name."""
       59│         try:
    →  60│             return self.registry[name]
       61│         except KeyError:
       62│             raise MissingFromRegistryError(name)
       63│
       64│     @property

The following error occurred when trying to handle this error:


  Stack trace:

  24  ~\AppData\Local\pypoetry\Cache\virtualenvs\test1-WxuTdTEg-py3.11\Lib\site-packages\cleo\application.py:327 in run
       325│
       326│             try:
     → 327│                 exit_code = self._run(io)
       328│             except BrokenPipeError:
       329│                 # If we are piped to another process, it may close early and send a

  23  ~\AppData\Local\pypoetry\Cache\virtualenvs\test1-WxuTdTEg-py3.11\Lib\site-packages\cleo\application.py:431 in _run
       429│             io.input.interactive(interactive)
       430│
     → 431│         exit_code = self._run_command(command, io)
       432│         self._running_command = None
       433│

  22  ~\AppData\Local\pypoetry\Cache\virtualenvs\test1-WxuTdTEg-py3.11\Lib\site-packages\cleo\application.py:438 in _run_command
       436│     def _run_command(self, command: Command, io: IO) -> int:
       437│         if self._event_dispatcher is None:
     → 438│             return command.run(io)
       439│
       440│         # Bind before the console.command event,

  21  ~\AppData\Local\pypoetry\Cache\virtualenvs\test1-WxuTdTEg-py3.11\Lib\site-packages\cleo\commands\base_command.py:119 in run
       117│         io.input.validate()
       118│
     → 119│         status_code = self.execute(io)
       120│
       121│         if status_code is None:

  20  ~\AppData\Local\pypoetry\Cache\virtualenvs\test1-WxuTdTEg-py3.11\Lib\site-packages\cleo\commands\command.py:62 in execute
        60│
        61│         try:
     →  62│             return self.handle()
[...]

PR #201, opened 2024-01-03
Add a Normalize Value Provider

zprobst:
The documentation in the PR explains the use case. Incidentally this makes some strides in improving the issue specified in #148 as well.


angelosantos4 (approved):
Huge, actually ran into this problem in some of the migration fixes.

PR #202, opened 2024-01-03
Feature/switch transformer

angelosantos4:
Switch transformers allow for the manipulation of records that do not arrive in the same format but can be distinguished by a consistent value.

Also bringing back ValueProjectionWithContext as an addon to ValueProjection seeing as it was more of an additive change.


nodestream/pipeline/transformers/transformer.py
```diff
@@ -129,3 +131,44 @@ def do_work_on_record(self, record: Any) -> Any:
...
+    ):
+        self.switch_on = ValueProvider.guarantee_value_provider(switch_on)
+        self.transformers = {
+            field_value: self.guarantee_transformer_from_file_data(transformer)
+            for field_value, transformer in cases.items()
+        }
+        self.default = (
+            self.guarantee_transformer_from_file_data(default) if default else None
+        )
+        self.normalization = normalization or {}
```
zprobst, line 163: Can you move the initialization logic to `from_file_data`.  The constructor should probably only take initialized class instances. For instance, check out: https://github.com/nodestream-proj/nodestream/blob/main/nodestream/pipeline/transformers/expand_json_field.py#L10

nodestream/pipeline/transformers/transformer.py
```diff
@@ -129,3 +131,44 @@ def do_work_on_record(self, record: Any) -> Any:
...
+    async def transform_record(self, record: Any):
+        context = ProviderContext.fresh(record)
+        key = self.switch_on.normalize_single_value(context, self.normalization)
+        transformer = self.transformers.get(key, self.default)
+        # If we don't have a transformer to handle the record we just yield the record (PassTransformer)
+        if transformer is None:
+            yield record
+        else:
+            async for result in transformer.transform_record(record):
+                yield result
```
zprobst, line 174: If we create a `PassTransformer`, we can default `self.default` to an instance of pass transformer and remove the branch from the code.

nodestream/pipeline/transformers/transformer.py
```diff
@@ -129,3 +131,44 @@ def do_work_on_record(self, record: Any) -> Any:
     @abstractmethod
     async def transform_record(self, record: Any) -> Any:
         raise NotImplementedError
+
+
+class SwitchTransformer(Transformer):
```
zprobst, line 136: Can you add usage documentation to our markdown docs.

zprobst (approved):

PR #206, opened 2024-01-08
fix resolving tags inside import resolver files

grantleehoffman:
BUGFIX
- Resolve tags inside !include tag files
- Add in the reading of declared targets from scope level in the project file

### DIFF
--- a/nodestream/project/pipeline_definition.py
+++ b/nodestream/project/pipeline_definition.py
@@ -1,7 +1,7 @@
 import os.path
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Dict
+from typing import Any, Dict, Set
 
 from ..file_io import LoadsFromYaml, SavesToYaml
 from ..pipeline import Pipeline, PipelineFile, PipelineInitializationArguments
@@ -29,7 +29,7 @@ class PipelineDefinition(IntrospectiveIngestionComponent, SavesToYaml, LoadsFrom
 
     name: str
     file_path: Path
-    targets: list[str] = None
+    targets: Set[str] = frozenset()
     annotations: Dict[str, Any] = field(default_factory=dict)
 
     @classmethod
@@ -58,7 +58,7 @@ class PipelineDefinition(IntrospectiveIngestionComponent, SavesToYaml, LoadsFrom
                     Optional("name"): str,
                     "path": os.path.exists,
                     Optional("annotations"): {str: Or(str, int, float, bool)},
-                    Optional("targets"): list[str],
+                    Optional("targets"): [str],
                 },
             )
         )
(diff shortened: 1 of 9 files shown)


zprobst (approved):
:shipit:

nodestream/project/pipeline_scope.py
```diff
@@ -26,26 +26,34 @@ def __init__(
...
+        targets: List[str] = [],
     ) -> None:
         self.persist = persist
         self.name = name
         self.config = config
         self.targets = targets
         self.pipelines_by_name: Dict[str, PipelineDefinition] = {}
         for pipeline in pipelines:
+            if pipeline.targets is not None:
+                pipeline.targets = list(set(pipeline.targets) | set(self.targets))
```
zprobst, line 38: This makes we think that maybe we should store the `target` as a `set` to begin with. What do you think?

nodestream/project/pipeline_scope.py
```diff
@@ -26,26 +26,34 @@ def __init__(
         pipelines: List[PipelineDefinition],
         persist: bool = True,
         config: ScopeConfig = None,
-        targets: List[str] = None,
+        targets: List[str] = [],
```
zprobst, line 29: I think the usage is technically safe here, but I think we should avoid the pitfall of using [mutable default arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

zprobst, line 29: I am thinking about the name `targets`.

To me I would expect `targets` to be a `List[Target]` not a `List[str]`. Three possible options:

- I get over it 
- Change `targets` to `target_names` 
- Resolve the actual names into `Target` objects and reference the `Target`s instead. 

I am very slightly leaning towards the third option, but it is more work. Thoughts?

nodestream/project/pipeline_definition.py
```diff
@@ -29,7 +29,7 @@ class PipelineDefinition(IntrospectiveIngestionComponent, SavesToYaml, LoadsFrom
 
     name: str
     file_path: Path
-    targets: str = None
+    targets: list[str] = None
```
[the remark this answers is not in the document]
zprobst: Follow up. I guess it was 3.10 so all good 👍

nodestream/project/pipeline_scope.py
grantleehoffman: We will change to Target types but agreed on this being part of a larger refactor since this isn't easily doable at the moment.

nodestream/project/pipeline_scope.py
```diff
@@ -26,26 +26,34 @@ def __init__(
...
+        targets: Set[str] = frozenset(),
     ) -> None:
         self.persist = persist
         self.name = name
         self.config = config
         self.targets = targets
         self.pipelines_by_name: Dict[str, PipelineDefinition] = {}
         for pipeline in pipelines:
+            if self.targets is not None:
+                pipeline.targets = set(pipeline.targets | self.targets)
```
zprobst, line 38: Minor thing here, but I think `set | set` should produce a `set` making calling `set` redundant.

zprobst (approved):

zprobst (approved):

zprobst (dismissed):

zprobst (approved):

PR #211, opened 2024-01-09
add option for disabling scope targets

grantleehoffman:
- Add pipeline option for disabling inherited targets


angelosantos4 (approved):
Pipelines 👍

PR #223, opened 2024-01-12
Fix Ingest Hooks and Writers

zprobst:
This PR introduces two fixes in `Writer` Logic:

- `Writers` did not handle flush globally, but rather required it to be handled locally. This pulls additional handling to the `Writer` base class. 
- Hooks were re-executed between batches.

### DIFF
--- a/nodestream/pipeline/writers.py
+++ b/nodestream/pipeline/writers.py
@@ -2,8 +2,8 @@ from abc import abstractmethod
 from logging import INFO, getLevelName, getLogger
 from typing import Any, AsyncGenerator
 
-from .step import Step
 from .flush import Flush
+from .step import Step
 
 
 class Writer(Step):
--- a/nodestream/pipeline/writers.py
+++ b/nodestream/pipeline/writers.py
@@ -3,6 +3,7 @@ from logging import INFO, getLevelName, getLogger
 from typing import Any, AsyncGenerator
 
 from .step import Step
+from .flush import Flush
 
 
 class Writer(Step):
@@ -17,12 +18,21 @@ class Writer(Step):
         self, record_stream: AsyncGenerator[Any, Any]
     ) -> AsyncGenerator[Any, Any]:
         async for record in record_stream:
-            await self.write_record(record)
+            if record is Flush:
+                await self.flush()
+            else:
+                await self.write_record(record)
             yield record
 
     @abstractmethod
     async def write_record(self, record: Any):
         raise NotImplementedError
+
+    async def flush(self):
+        pass
+
+    async def finish(self):
+        await self.flush()
 
 
 class LoggerWriter(Writer):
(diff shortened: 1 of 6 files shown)


nodestream/pipeline/writers.py
```diff
@@ -17,13 +18,22 @@ async def handle_async_record_stream(
         self, record_stream: AsyncGenerator[Any, Any]
     ) -> AsyncGenerator[Any, Any]:
         async for record in record_stream:
-            await self.write_record(record)
+            if record is Flush:
```
yasonk, line 21: Just to understand. The extractor generates Flush as an EOF marker of sorts. Then the writer recognizes it and tells  the ingestion strategy to flush. Initial thought is why not put this conditional on the ingestion strategy and make Flush an ingestible?

If my understanding is correct, flushing is a side effect of an ingestion strategy that is enabled by the injection of the EOF marker by the extractor. In this case writer is an auxiliary when it doesn't need to be.

angelosantos4: My interpretation of this change was that the writer would be able to flush to a downstream step like another writer or another ingestion component. Like suppose there is something complicated like:

Extractor:
Transformer 1:
Writer 1:
ConcurrentTransformer 2: 
Ingestion:
Writer 2:

Something like the flush would be necessary for ConcurrentTransformer 2 to know that it shouldn't expect any new data from the upstream service and shouldn't spawn new processes.
Or if we were just writing to writer 1 and writer 2, it lets it know that it shouldn't wait for the pending_records to reach the batch size limit. Like pending_records = 500 < 1000, the writer will just write what we have.

zprobst: @angelosantos4 has essentially the right idea. Does that explanation make sense @yasonk ?

yasonk: @angelosantos4 @zprobst I think I understand, but my comment stems from the lower level code. Where the writer simply calls `ingestion_strategy.flush()`.
I guess my understanding wrong that flushing is a feature of the ingestion_strategy.

yasonk: I guess the code I was looking at already sort of exists, but not sure how it is hooked up:
https://github.com/nodestream-proj/nodestream/blob/a2ece8a6741cf4e7a488ef3d21149d8173e6ba31/nodestream/databases/writer.py#L41

yasonk (approved):

nodestream/pipeline/writers.py
zprobst: Capturing an offline conversation for the record:

Essentially this conversation boiled down to:

- Handling of `Flush` did not belong in `GraphDatabaseWriter`
- Q: Does it belong in `Writer` or `IngestStrategy`?
- A: It has to belong in `Writer` in the current state because `Writer` subclasses don't strictly operate on `ingestible` things.

PR #226, opened 2024-01-17
Handle null values and broken paths in ExpandJsonField

zprobst:
- Handles null values in path parts 
- Handles invalid (for the object) paths.

### DIFF
--- a/nodestream/pipeline/transformers/expand_json_field.py
+++ b/nodestream/pipeline/transformers/expand_json_field.py
@@ -16,15 +16,25 @@ class ExpandJsonField(Transformer):
     def __init__(self, path: List[str]) -> None:
         self.path = path
 
+    def dig_to_path(self, record: JsonLikeDocument, path: List[str]):
+        value_at_path = record
+        for path_segment in path:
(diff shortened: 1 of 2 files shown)


nodestream/pipeline/transformers/expand_json_field.py
```diff
@@ -19,7 +19,12 @@ def __init__(self, path: List[str]) -> None:
     async def transform_record(self, record: JsonLikeDocument):
         item = record
         for path_segment in self.path[:-1]:
-            record = item[path_segment]
+            item = item.get(path_segment, None)
```
yasonk, line 22: Not sure why `item = record` above is needed. The interlaced use of `record` and `item` is extremely confusing.

For example, this line could be:
`path_segment= record.get(path_segment, None)`

then below:
`last_segment_value = path_segment.get(last_segment, None)

yasonk: I guess that won't work... let me think what else can be done

yasonk (approved):

PR #228, opened 2024-01-17
added aiobotocore library

leszek-bq:


leszek-bq:
Sorry did not mean to submit a pull request on the main project

PR #233, opened 2024-01-23
fix loading plugin config

grantleehoffman:
## BugFix
- Load plugin configurations and plugin scope add.

### DIFF
--- a/nodestream/project/project.py
+++ b/nodestream/project/project.py
@@ -211,10 +211,8 @@ class Project(
 
     def add_plugin_scope(self, name: str, plugin: PluginConfiguration):
         project_plugin_configuration = self.plugins_by_name.get(name)
-
         if project_plugin_configuration:
-            plugin.config = self.plugins_by_name.get(name).config
-            plugin.update_pipeline_configurations(project_plugin_configuration)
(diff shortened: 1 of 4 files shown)


ccloes (approved):

nodestream/project/project.py
```diff
@@ -213,6 +213,7 @@ def add_plugin_scope(self, name: str, plugin: PluginConfiguration):
         project_plugin_configuration = self.plugins_by_name.get(name)
 
         if project_plugin_configuration:
+            plugin.config = self.plugins_by_name.get(name).config
```
zprobst, line 216: Minimally, I think we can make this adjustment. 

```suggestion
            plugin.config = project_plugin_configuration.config
            plugin.update_pipeline_configurations(project_plugin_configuration)
```

However, I think I'd recommend moving this line down into `update_pipeline_configurations` and potentially renaming it so that `Project` does not have to be concerned with the inner data model of `PluginConfiguration`.

grantleehoffman: whoops yeah wasn't done with the refactoring, did excatly that

zprobst (approved):

PR #236, opened 2024-01-24
S3 extractor fix

angelosantos4:


nodestream/pipeline/extractors/files.py
```diff
@@ -58,7 +58,7 @@ def read_file_from_handle(self, fp: StringIO) -> Iterable[JsonLikeDocument]:
 
 class JsonFileFormat(SupportedFileFormat, alias=".json"):
     def read_file_from_handle(self, fp: StringIO) -> Iterable[JsonLikeDocument]:
-        return [json.load(fp)]
+        return [json.loads(fp)]
```
angelosantos4, line 61: This class method expects a string as input, so it must use json.loads()

nodestream/pipeline/extractors/stores/aws/s3_extractor.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         self.logger = getLogger(__name__)
 
     def get_object_as_io(self, key: str) -> StringIO:
-        return self.s3_client.get_object(Bucket=self.bucket, Key=key)["Body"]
+        return self.s3_client.get_object(Bucket=self.bucket, Key=key)["Body"].read().decode('utf-8')
```
angelosantos4, line 45: self.s3_client.get_object(Bucket=self.bucket, Key=key)["Body"] is a botocore.response.StreamingBody type,
This has a read() method that turns it into a Bytes type, and the decoding turns it into a string.

nodestream/pipeline/extractors/files.py
zprobst: This does not expect a string as input, it expects a `StringIO`.

nodestream/pipeline/extractors/stores/aws/s3_extractor.py
zprobst: See comment above, we do not want or need to read to a string.

PR #259, opened 2024-02-13
Adding the new command option for k8s

angelosantos4:


nodestream/cli/commands/shared_options.py
```diff
@@ -19,3 +19,10 @@
 MANY_PIPELINES_ARGUMENT = argument(
     "pipelines", "the names of the pipelines", multiple=True, optional=True
 )
+
+DEVELOPMENT_ENVIRONMENT_OPTION = option(
+    "development",
+    "g",
+    "Run development settings. Cronjobs are suspended, perpetuals are scaled to 0 unless the debug annotation is specified.",
+    flag=True,
+)
```
zprobst, line 28: There seems to be no reason why this was added to nodestream core aside from its use in https://github.com/nodestream-proj/nodestream-plugin-k8s/pull/6/files. Is that right?

If so, I think it would be better to move this to that PR.

nodestream/pipeline/extractors/ttls.py
```diff
@@ -7,13 +7,21 @@
...
+        configurations: Iterable[Dict[str, Any]],
+        global_expiry_in_hours: int = None,
     ) -> None:
         self.graph_object_type = GraphObjectType(graph_object_type)
         self.configurations = configurations
+        self.global_expiry_in_hours = global_expiry_in_hours
 
     async def extract_records(self):
         for config in self.configurations:
+            # Only when global_expiry_in_hours is set, will we have an overwrite of all ttl's
```
zprobst, line 21: I think we need to add an explanation similar to this to the documentation on TTLs.

nodestream/pipeline/extractors/ttls.py
```diff
@@ -7,13 +7,21 @@
...
+            if self.global_expiry_in_hours:
+                config["expiry_in_hours"] = self.global_expiry_in_hours
```
zprobst, line 23: We also need to add a test for this feature.

nodestream/pipeline/extractors/ttls.py
```diff
@@ -7,13 +7,21 @@
...
     def __init__(
-        self, graph_object_type: str, configurations: Iterable[Dict[str, Any]]
+        self,
+        graph_object_type: str,
+        configurations: Iterable[Dict[str, Any]],
+        global_expiry_in_hours: int = None,
     ) -> None:
         self.graph_object_type = GraphObjectType(graph_object_type)
         self.configurations = configurations
+        self.global_expiry_in_hours = global_expiry_in_hours
```
zprobst, line 17:
```suggestion
        override_expiry_in_hours: int = None,
    ) -> None:
        self.graph_object_type = GraphObjectType(graph_object_type)
        self.configurations = configurations
        self.override_expiry_in_hours = override_expiry_in_hours
```

`global`, in my mind, implies that it is set for all types when it actuality only set for types that are provided.

zprobst (changes requested):

zprobst (approved):

PR #263, opened 2024-02-20
Timestamp Caching

zprobst:
Was playing around with some rust code over the weekend. On a hunch, I had a feeling that getting datetimes from pandas was going to be slow and alas... it was. This implements a time cache so that we don't get Pandas datetimes 1000s of times a second. 

# Current Performance
old

# After Optimization
new


nodestream/model/graph_objects.py
```diff
@@ -18,22 +22,42 @@ def update(self, other: "DeduplicatableObject"):
...
+# basically the same, there is little value in getting the current time over
+# and over again. So we cache the current time for a short period of time.
+# Profiling this generally shows a 5-15% speedup in the time it takes to
+# ingest a large number of nodes and relationships (Excluding database time).
+
+
+@lru_cache(maxsize=128)
+def _get_cached_timestamp(time_hash: int) -> Timestamp:
+    del time_hash  # Unused; just needed to cache the result
+    return Timestamp.utcnow()
```
yasonk, line 37: I wonder if maybe we can just use datetime.utcnow() and then print it with the right format?

nodestream/model/graph_objects.py
```diff
@@ -18,22 +22,42 @@ def update(self, other: "DeduplicatableObject"):
...
+
+
+def get_cached_timestamp(max_age_in_seconds: int = 2) -> Timestamp:
+    # calling time.time() is relatively fast, but getting a Timestamp object
+    # is relatively slow.
+    ttl_hash = round(time.time()) / max_age_in_seconds
```
yasonk, line 43: not sure I understand what max_age_in_seconds does...
If time = 3
ttl_hash = 1.5

If time = 4
ttl_hash = 2

From what I understand then you will get a different timestamp?

tests/unit/model/test_graph_objects.py
```diff
@@ -61,3 +64,16 @@ def test_relationship_into_ingest():
 def test_node_key_validity(keys, expected):
     node = Node("Person", keys)
     assert_that(node.has_valid_id, equal_to(expected))
+
+
+def test_get_cached_timestamp():
+    # The first and second timestamp should be the same if called in quick
+    # succession.
+    first = get_cached_timestamp()
```
yasonk, line 72: I think the test would be better if we could mock or inject the time values to test different edge cases.

For example, this test works for two timestamps close together and also 2 seconds apart.
But what if I'm not sure what happens when timestamps come down on different sides of the break. Ex: 1.5 and 1.6 or 1.9 and 2.0.

yasonk: Also it could add brittleness. For example, if there is a bug in the code making the test fail only at a certain time, you wouldn't be able to reproduce it, the build would break, but also we probably don't care if 1 out of 1000 timestamps doesn't cache the same as the rest...

nodestream/model/graph_objects.py
zprobst: Ah good catch. The reason this doesn't make sense is the round is supposed to be around the whole operation `time.time() / max_age_in_seconds`. In your scenario, 

If time = 3
ttl_hash = 1 

if time = 4 
ttl_hash=2

tests/unit/model/test_graph_objects.py
zprobst: Yeah... fair point... I'll take another crack at this.

nodestream/model/graph_objects.py
zprobst: Not sure I fully understand

yasonk: I wasn't sure how this was used. But if we can use `datetime.utcnow().timestamp()` AND that is faster than Timestamp.utcnow(), maybe we could use that.

Stackoverflow is telling me that `datetime.utcnow()` is magnitudes faster

nodestream/model/graph_objects.py
yasonk: btw round 1.5 = 2. But yes that makes more sense.

yasonk (approved):

PR #264, opened 2024-02-21
fix jsonl line parsing

grantleehoffman:
- Fix error when parsing jsonl documents from s3 extractor.
- iterator of `StringIO` type returns type `bytes`, using iterator of `StringIO.readlines()` returns type `str`
- File reader open as bytes and specify reader in file handlers


angelosantos4:
I think maybe also add a test for jsonl file types in the S3 Extractor test files:
tests/unit/pipeline/extractors/stores/aws/test_s3_extractor.py

That replicates the JSON decode error when these changes are undone.

zprobst (approved):

PR #266, opened 2024-02-22
Correct batchsize to be batchSize

zprobst:
This was a typo from the query. This value is being ignored because of casing. 

The reason this value should can be set is that large transactions can cause major turmoil when it its updating with large transactions of relationships (Especially against large data sets). 

See here in the docs:
https://neo4j.com/docs/apoc/current/overview/apoc.periodic/apoc.periodic.iterate/


grantleehoffman (approved):

ccloes (approved):

grantleehoffman:
Good Catch! 🎉

yasonk (approved):

PR #268, opened 2024-02-29
0.12

zprobst:
Working Branch for Nodestream `0.12`


nodestream/cli/operations/run_project_cookiecutter.py
```diff
@@ -15,13 +15,17 @@ def __init__(self, project_name, project_path, database):
...
         cookiecutter(
             PROJECT_COOKIECUTTER_URL,
             no_input=True,
             output_dir=self.project_path,
             extra_context={
                 "project_name": self.project_name,
                 "database": self.database,
-                "nodestream_version": get_version(),
+                "nodestream_version": version,
+                "nodestream_semver_non_patch_version": semver_non_patch_version,
```
angelosantos4, line 29: What does this code portion do again? I understand this part is relevant to the Neptune plugin introduction.

zprobst: `nodestream_semver_non_patch_version` would equal `0.12` for the upcoming release. It will be used by the template to get the version `0.12` for the associated plugin. Without this, we'd only have `nodestream_version` which would equal something like `0.12.12` down the road. We don't want to release patch versions across all dbs every time we release the core library.

angelosantos4: Ahh okay makes sense, so it just takes 0.a out of 0.a.b. So only when a major 0.13 comes out well have to rev the plugins.

docs/technology.md
```diff
@@ -18,15 +18,15 @@ Parallelism deals with multiple operations executing simultaneously while concur
 
 `asyncio` operates on the observation that in some systems, the cpu is spending a lot of time _blocking_ or waiting for IO operations such as network calls, file reads, and the like to finish before it can continue operating. 
 To make better use of the CPU, `asyncio` introduces a series of language features that facilitate defining tasks and signaling when they need to wait. The language runtime can the “swap in” a different task that it can work on that’s not (currently) IO bound. 
-In workloads where there is an abundant amout of IO, the overhead of managing these tasks is worth it because  we are able to make better use CPU time that would otherwise be wasted.
+In workloads where there is an abundant amount of IO, the overhead of managing these tasks is worth it because  we are able to make better use CPU time that would otherwise be wasted.
```
grantleehoffman, line 21:
```suggestion
In workloads where there is an abundant amount of IO, the overhead of managing these tasks is worth it because  we are able to make better use of CPU time that would otherwise be wasted.
```

angelosantos4 (dismissed):
First pass review looks good. Everything seems organized and the code runs on my end.
Will still need to play with the migration changes to fully understand it.

grantleehoffman (approved):

zprobst:
🎆

PR #280, opened 2024-03-26
Modified the normalizer to handle single values properly + Neo4j Writer Logs URI

angelosantos4:
- Fixing a bug with the normalize value provider that crashes when the data returns None. 
- Adding database uri for logging. This can be used to specify logs between different database targets.
- Increase the timeout window for error collecting tests for more deterministic tests.

### DIFF
--- a/nodestream/databases/neo4j/query_executor.py
+++ b/nodestream/databases/neo4j/query_executor.py
@@ -25,7 +25,6 @@ class Neo4jQueryExecutor(QueryExecutor):
         chunk_size: int = 1000,
         execute_chunks_in_parallel: bool = True,
         retries_per_chunk: int = 3,
-        alias: str = "neo4j-default",
     ) -> None:
         self.driver = driver
         self.ingest_query_builder = ingest_query_builder
@@ -35,7 +34,6 @@ class Neo4jQueryExecutor(QueryExecutor):
         self.chunk_size = chunk_size
         self.execute_chunks_in_parallel = execute_chunks_in_parallel
         self.retries_per_chunk = retries_per_chunk
-        self.alias = alias
 
     async def execute_query_batch(self, batch: QueryBatch):
         await self.execute(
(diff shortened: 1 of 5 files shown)


nodestream/pipeline/value_providers/normalizer_value_provider.py
```diff
@@ -13,7 +13,11 @@ def __init__(self, using: str, data: ValueProvider):
         self.data = data
 
     def single_value(self, context: ProviderContext) -> Any:
-        return next(self.many_values(context))
+        values = list(self.many_values(context))
+        if values:
+            return next((value for value in values))
+        else:
+            return None
```
zprobst, line 20:
```suggestion
        return next(self.many_values(context), default=None)
```

I think this is a cleaner solution

nodestream/databases/neo4j/query_executor.py
```diff
@@ -25,6 +25,7 @@ def __init__(
         chunk_size: int = 1000,
         execute_chunks_in_parallel: bool = True,
         retries_per_chunk: int = 3,
+        alias: str = "neo4j-default",
```
zprobst, line 28: Why introduce another value here? Why not rely on the URI and log it out?  See [here](https://github.com/nodestream-proj/nodestream/blob/280298babb589743aa544074a3bd6a45cdf44001/nodestream/databases/neo4j/query_executor.py#L94).

angelosantos4: Figured we wouldn't want the uri in the logs, but I can add it into the log message instead.

nodestream/pipeline/value_providers/normalizer_value_provider.py
angelosantos4: Yeah thats way cleaner didn't even notice that was an option.

zprobst (approved):

PR #284, opened 2024-04-04
Fix `Target` and `Schema` bugs

zprobst:
This PR Fixes two issues:

1. If you have two pipelines; one that says a property is a key and another that says it only knows the property as a property, the order of operations takes precedence. This occurs when using a match only clause. If we know something is a key, it needs to be defined as such. Same thing with indexes. 

2. Currently we are passing writer configurations (such as `batch_size`) through the target. Currently these values are getting passed as part of the the connector configuration directly to the writer that is able to sort through these arguments and use them. However, with the addition of the `Migrator`, these configuration values are treated as invalid. This PR shuffles the writer configuration values as well the connector specific values away from one another and only sends the writer arguments when appropriate.

### DIFF
--- a/nodestream/databases/database_connector.py
+++ b/nodestream/databases/database_connector.py
@@ -15,7 +15,7 @@ class DatabaseConnector(ABC, Pluggable):
 
     @classmethod
     def from_database_args(
-        cls, database: str = "neo4j", **database_args
+        cls, database: str = "null", **database_args
     ) -> "DatabaseConnector":
         DatabaseConnector.import_all()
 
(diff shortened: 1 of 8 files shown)


nodestream/databases/database_connector.py
```diff
@@ -15,7 +15,7 @@ class DatabaseConnector(ABC, Pluggable):
 
     @classmethod
     def from_database_args(
-        cls, database: str = "neo4j", **database_args
+        cls, database: str = "null", **database_args
```
zprobst, line 18: This is is just so the null connector is used instead of a default that may or may not be installed.

yasonk (approved):

angelosantos4 (approved):

PR #289, opened 2024-04-10
Feat/add allow create flag to source nodes

yasonk:
This a WIP. Do not review


yasonk:
Will use the v2 branch instead

PR #292, opened 2024-04-11
Feat/add allow create flag to source nodes v2

yasonk:
When source node of a pipeline is fully controlled by another pipeline, we may not want to create new nodes.

For example, if we already have a pipeline that creates all the Persons from our employee database, a new pipeline that attaches address to Persons may want avoid creating "Person" nodes simply because the address for that person already exits.


nodestream/model/relationship_draft.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Optional
+
+from .creation_rules import NodeCreationRule, RelationshipCreationRule
+from .graph_objects import Node, Relationship, RelationshipWithNodes
+
+LOGGER = getLogger(__name__)
+
+
+@dataclass(slots=True)
+class RelationshipDraft:
```
yasonk, line 12: Do we need this class? Maybe we can just add unfinished `RelationshipWithNodes` to a separate list and then the finalize method simply adds the to_node and the from_node data?

zprobst: I've been writing some rust recently and they love this [type state](https://cliffle.com/blog/rust-typestate/) pattern. I think this is what was in my head when we were talking about this. Mostly I think this criteria applies here:

> Operations on an object (such as methods or functions) that are only available when the object is in certain states

Now arguably that pattern doesn't really map the same way in python. But I do prefer this separation of methods regardless.

nodestream/interpreting/interpretations/relationship_interpretation.py
```diff
@@ -153,8 +153,11 @@ def interpret(self, context: ProviderContext):
 
     def find_relationship(self, context: ProviderContext) -> Relationship:
         rel = Relationship(type=self.relationship_type.single_value(context))
-        rel.key_values.apply_providers(
-            context, self.relationship_key, self.key_normalization
+        relationship_key_property_mapping = PropertyMappingFromDict(
+            self.relationship_key
```
yasonk, line 157: relationship_key is used in `expand_relationship_schema`:
```
relationship_schema.add_keys(self.relationship_key)
```

tests/unit/model/test_desired_ingestion.py
```diff
@@ -65,3 +70,73 @@ async def test_run_ingest_hooks(desired_ingestion, mocker):
             mocker.call(desired_ingestion.hook_requests[1]),
         ]
     )
+
+
+def test_relationship_is_finalized_after_source_node_added(
```
zprobst, line 75: Note to our future selves:
 Its a little sad that these tests have such a large structure to them. Maybe we need to take a look into something like [Faker](https://faker.readthedocs.io/en/master/).

nodestream/model/desired_ingestion.py
```diff
@@ -64,6 +65,37 @@ async def ingest(self, strategy: "IngestionStrategy"):
...
+        """
+        for relationship in self.relationships:
+            if relationship.outbound:
+                relationship.from_node = self.source
+                relationship.from_side_node_creation_rule = (
+                    self.source_node_creation_rule
+                )
+            else:
+                relationship.to_node = self.source
+                relationship.to_side_node_creation_rule = self.source_node_creation_rule
```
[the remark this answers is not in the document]
yasonk: Added some unit tests for this: test_inbound_relationship_is_finalized_after_source_node_added

zprobst (approved):

PR #304, opened 2024-05-14
Refactoring the Pipeline Runtime

### LINKED ISSUE #187: [REQUEST] Redesign the Pipeline Runtime
Currently, there are a number of limitations and issues associated with the data model for the core Internals of the Pipeline runtime and the step interface. Here is a (non-exhaustive) list.

1. The `Step` class is currently defined by a function that takes a `AsyncGenerator` and returns an `AsyncGenerator`.  This leads to some complex code in the `StepExecutor` class to do [gymanastics to conform to that interface](https://github.com/nodestream-proj/nodestream/blob/420ed5b755e662b34682d76af91167ec2b1fae22/nodestream/pipeline/pipeline.py#L128).
2. `Step`s are also constrained in the hooks that they can connect with. For instance, there is no step hook that a step can rely on to perform clean up operations or other work when the pipeline is complete. We should consider prior art in this area specifically. What hooks are provided to steps in other pipeline models? What works for `nodestream`? 
3. The current runtime cannot scale beyond one thread or one process which limits the performance capabilities of this model. This is partially because of the first issue mentioned and partially do to other limitations in the way the runtime is written. 

The specifics of the solution are unclear. However, it is clear that a comprehensive cleanup, rewrite, or refactor is required to support these chief limitations.

zprobst:
NOTE: The target branch for this is 0.13. 

This change introduces a better seperation of responsibility between the different aspects of the pipeline runtime. For instance, there are now specialized input and output classes that handle waiting for streaming data between steps as opposed to jamming things into the step context. 

This allows steps access to the context so that they can individually emit messages through context. 

Closes #187


zprobst:
Coverage drop due to removal of covered lines.

nodestream/pipeline/pipeline.py
```diff
@@ -1,280 +1,195 @@
...
+        await self.start_step()
+        await self.drive_step()
+        await self.output.done()
+        self.input.done()
+        await self.stop_step()
+        self.context.debug(f"Finished step {self.step.__class__.__name__}")
 
-        return "Exceptions in Pipeline:\n" + message
 
+class PielineOutput:
```
angelosantos4, line 88: Misspelling here for PipelineOutput

nodestream/pipeline/pipeline.py
```diff
@@ -1,280 +1,195 @@
...
+        # Create the executors for the steps in the pipeline. The executors
+        # will be used to run the steps concurrently. The steps are created in
+        # reverse order so that the output of each step is connected to the
+        # input of the next step.
+        for index, step in reversed(list(enumerate(self.steps))):
+            context = StepContext(index, reporter)
+            current_input, next_output = channel(self.step_outbox_size)
+            exec = StepExecutor(step, current_input, current_output, context)
+            current_output = next_output
+            executors.append(exec)
```
angelosantos4, line 184: Just a thought about this one, I'm assuming this implies that the context for each step will be initialized in reverse order (Extractor 3, Transformer 2, Ingester 1, Writer 0). It kind of seems odd in terms of identifying the ordering of events when looking at the context logs. I believe this piece can be unreversed with the pipeline output being initialized at the end. Also I think this is a good oppertunity for the StepContext to obtain metadata about the step name from the class info here.

nodestream/pipeline/pipeline.py
```diff
@@ -1,280 +1,195 @@
...
+                        return
+
+            async for record in self.step.emit_outstanding_records():
+                if not await self.emit_record(record):
+                    return
+
+            self.context.debug(f"Step {self.step.__class__.__name__} finished emitting")
+        except Exception as e:
+            self.context.report_error(
+                f"Error running step {self.step.__class__.__name__}",
```
angelosantos4, line 73: Something like this is nice when parsing logs as messages, but it would also be nice to have the class name and index in dictionary form {step_name: "__class__.__name__", index: 1}. With something like this we wouldn't have to regex the messages to isolate something like errors by step.

nodestream/pipeline/transformers/transformer.py
```diff
@@ -54,83 +57,64 @@ def __init__(
...
+            # If we don't have capacity, yield completed tasks.
+            # Once there is nothing remaining to yield from these tasks,
+            # yield the processor until capacity is available.
+            submitted = False
+            while not submitted:
+                if len(self.pending_tasks) < self.maximum_pending_tasks:
+                    task = self.thread_pool.submit(self.do_work_on_record, record)
+                    self.pending_tasks.append(task)
+                    submitted = True
+                async for result in self.drain_completed_tasks():
```
angelosantos4, line 99: I think using something like `async def emit_outstanding_records(self): ` here is better since drain_completed_tasks() does not yield the processor at any point. Unless this piece is intended on blocking (so something like the extractor does not put more records with the new channel update.) this should yield the processor so the pipeline doesn't get stuck waiting for results in this blocking section.

nodestream/pipeline/pipeline.py
zprobst: `Pieline`.... totally intentional

nodestream/pipeline/pipeline.py
zprobst: Really good call.

nodestream/pipeline/pipeline.py
zprobst: Yeah thats miss on my part. `Extractor` from your example should be `0`. I'll math that.

nodestream/pipeline/transformers/transformer.py
zprobst: Hmm yes... I see what you mean. this is mend to be calling `emit_outstanding_records` which does yield.

nodestream/pipeline/channel.py
```diff
@@ -0,0 +1,163 @@
...
+            if the channel is full and the object was not put in the channel.
+        """
+        try:
+            await wait_for(self.queue.put(obj), timeout=CHANNEL_TIMEOUT)
+            return True
+        except (TimeoutError, asyncio.TimeoutError):
+            return False
+
+
+class StepOuput:
```
angelosantos4, line 74: Just noticed, misspelled Output.

nodestream/pipeline/transformers/transformer.py
```diff
@@ -54,83 +57,65 @@ def __init__(
...
+            self.logger.debug(
+                "Drained %s completed tasks, %s pending tasks remain",
+                tasks_drained,
+                len(self.pending_tasks),
+            )
+
+    async def process_record(self, record: Any, _) -> Any:
+        if record is Flush:
+            async for outstanding in self.drain_completed_tasks():
+                yield outstanding
```
angelosantos4, line 88: Outstanding case for a Flush being passed as a record. In the unit test this could just be added as one of the elements in the input generator.

tests/unit/pipeline/transformers/test_transformer.py
```diff
@@ -62,44 +53,15 @@ async def input_record_stream():
...
-            transformer.queue_size -= 1
-            should_continue = not transformer.done
-            await asyncio.sleep(0)
-
-    await asyncio.gather(transform(), downstream_client())
-
-
 @pytest.mark.asyncio
 async def test_normal_concurrent_transformer_passes_processor():
     items = list(range(ITEM_COUNT))
```
angelosantos4, line 58:
```suggestion
    items = list(range(ITEM_COUNT)) + [Flush]
```

angelosantos4 (approved):

PR #305, opened 2024-05-28
Search for matches anywhere in the string

yasonk:
Currently Regex Value Provider automatically anchors to the begging of the string. That means that searching the string for a specific pattern would always require having .* in the beginning.
Ex:
Given:
text = "I am Kevin"

1) regex = "Kevin" - Would find no matches
2) regex = ".*Kevin" - would find matches

Since this is not a common constraint on regexes, using python `re.search()` method will result in matches being found for both regexes.

### DIFF
--- a/tests/unit/pipeline/value_providers/test_regex_value_provider.py
+++ b/tests/unit/pipeline/value_providers/test_regex_value_provider.py
@@ -11,7 +11,7 @@ from ...stubs import StubbedValueProvider
 def subject_with_named_groups():
     return RegexValueProvider(
         regex="(?P<name>[a-z]+) (?P<age>[0-9]+)",
-        data=StubbedValueProvider(["I am john 42", "I am jane 32"]),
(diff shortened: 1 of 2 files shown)


tests/unit/pipeline/value_providers/test_regex_value_provider.py
```diff
@@ -11,7 +11,7 @@
 def subject_with_named_groups():
     return RegexValueProvider(
         regex="(?P<name>[a-z]+) (?P<age>[0-9]+)",
-        data=StubbedValueProvider(["john 42", "jane 32"]),
+        data=StubbedValueProvider(["I am john 42", "I am jane 32"]),
```
zprobst, line 14: Should we keep one of these the same so we can test that it matches if its the whole string?

zprobst (dismissed):
Looks good but see the comment and let me know what you think?

zprobst (approved):

PR #316, opened 2024-06-11
Adding the scope to all log messages.

angelosantos4:
This addition adds the scope to the log messages sent from pipelines. This way we can differentiate between these pipelines when doing log analysis using tools such as Splunk.

### DIFF
--- a/nodestream/project/run_request.py
+++ b/nodestream/project/run_request.py
@@ -17,6 +17,7 @@ class RunRequest:
     """
 
     pipeline_name: str
+    pipeline_scope: str
     initialization_arguments: PipelineInitializationArguments
     progress_reporter: PipelineProgressReporter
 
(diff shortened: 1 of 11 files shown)


nodestream/project/run_request.py
```diff
@@ -17,6 +17,7 @@ class RunRequest:
     """
 
     pipeline_name: str
+    pipeline_scope: str
```
zprobst, line 20: At first glance this seems but fine. It definitely gets the job done. But I'm troubled by `RunRequest`  getting `pipeline_scope`. 

If the request is meant to model what the user asked to have run, then have a `pipeline_scope` fields feels like it should be a filter or specifier and not just metadata that is being carried forward.  Feels like this is emblematic of an encapsulation problem.

zprobst (approved):

PR #318, opened 2024-06-17
Add sqs queue extractor

grantleehoffman:
Adds Queue extractor and SQS connector for processing queue messages from a SQS queue.
Has property `delete_after_read` which defaults to `True`. This will delete queue message batches after successfully reading to pipeline from extractor step.

Future Feature: Add in tracking of message processing through pipeline and delete once all writes are finished for each message.


nodestream/pipeline/extractors/queues/sqs.py
```diff
@@ -0,0 +1,86 @@
...
+                break
+        return results
+
+    async def get_next_messsage_batch(self):
+        loop = asyncio.get_running_loop()
+        msgs = await loop.run_in_executor(None, self.get_message_batch)
+        messages = msgs.get("Messages")
+        if messages is None:
+            return None
+        return self.process_messages(messages)
```
zprobst, line 72: I am None and boolean averse. I think we can achieve the same behavior and simply out this if statement by having `msgs.get("Messages")` default to an empty list with `msgs.get("Messages", [])` and then when we check if `if messages is None`, simply check if `len(messages) ==. 0`. This removes the conditional and changes the type from list | none to just list. We can also then drop the continue simplifying the loop.

nodestream/pipeline/extractors/queues/extractor.py
```diff
@@ -0,0 +1,88 @@
...
+    entrypoint_name = "queue_connectors"
+
+    @abstractmethod
+    async def poll(self) -> Iterable[Any]:
+        raise NotImplementedError
+
+
+@QUEUE_OBJECT_FORMAT_SUBCLASS_REGISTRY.connect_baseclass
+class QueueRecordFormat(Pluggable, ABC):
+    entrypoint_name = "record_formats"
```
zprobst, line 27: This seems to use the same name as the stream record formats.

nodestream/pipeline/extractors/queues/extractor.py
```diff
@@ -0,0 +1,88 @@
...
+class QueueConnector(Pluggable, ABC):
+    entrypoint_name = "queue_connectors"
+
+    @abstractmethod
+    async def poll(self) -> Iterable[Any]:
+        raise NotImplementedError
+
+
+@QUEUE_OBJECT_FORMAT_SUBCLASS_REGISTRY.connect_baseclass
+class QueueRecordFormat(Pluggable, ABC):
```
zprobst, line 26: I get `QueueConnector` and `QueueExtractor ` being a separate from stream extractor. By I am curious of `QueueRecordFormat` needs to be separate. The way I am thinking about it, a message is a message and on some level if it came from a queue based system or a stream based system is somewhat immaterial. I am curious what led you to separating this out and what your thoughts are here.

yasonk:
@grantleehoffman Should this have an update in the documentation? Maybe worth adding it to the docs here:
https://nodestream-proj.github.io/docs/docs/reference/extractors/

nodestream/pipeline/extractors/queues/extractor.py
```diff
@@ -0,0 +1,88 @@
...
+
+class QueueExtractor(Extractor):
+    """A QueueExtractor implements the standard behavior of polling data from a stream.
+
+    The QueueExtractor requires both a QueueConnector and a QueueRecordFormat to delegate
+    to for the actual polling implementation and parsing of the data, respectively.
+    """
+
+    @classmethod
+    def from_file_data(cls, connector: str, record_format: str, **connector_args):
```
zprobst, line 47: It is possible to add a test case for this? It looks like it may be tricky and that may have been why you omitted it in which case I am okay.

nodestream/pipeline/extractors/queues/sqs.py
```diff
@@ -0,0 +1,86 @@
...
+        cls,
+        queue_url: str,
+        message_system_attribute_names: Optional[list[str]] = ["All"],
+        message_attribute_names: Optional[list[str]] = ["All"],
+        max_batch_size: int = 10,
+        delete_after_read: bool = True,
+        max_batches: int = 10,
+        **aws_client_args,
+    ):
+        return cls(
```
zprobst, line 22: It is possible to add a test case for this? It looks like it may be tricky and that may have been why you omitted it in which case I am okay.

nodestream/pipeline/extractors/queues/extractor.py
grantleehoffman: I don't think it really needs to be I'll update

nodestream/pipeline/extractors/queues/extractor.py
```diff
@@ -0,0 +1,88 @@
+import json
```
yasonk, line 1: Maybe overkill, but could be beneficial to add some design diagrams to the developer reference: https://nodestream-proj.github.io/docs/docs/category/developer-reference/

docs/docs/reference/extractors.md
```diff
@@ -55,6 +55,54 @@ set the `record_format` to be `json` in the `StreamExtractor` configuration. For
      record_format: json
 ```
 
+## `QueueExtractor`
```
zprobst, line 58: Same thing for here: 

https://github.com/nodestream-proj/docs/blob/main/docs/reference/extractors.md

grantleehoffman: I'll create a pr for the public docs after we finalize this pr.

nodestream/pipeline/extractors/queues/extractor.py
grantleehoffman: Added

grantleehoffman:
I will add an update to the public docs once these changes have gone through review and are finalized.

zprobst (approved):

PR #323, opened 2024-06-26
DynamoDB Extractor

angelosantos4:


nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+from ...credential_utils import AwsClientFactory
+from ...extractor import Extractor
+
+
+class DynamoObject:
+    """
+        Dynamo objects generally come in the form {"attribute_type": value}
+        Example:
+            {"S", "some_string"}
+            {"N", "12345"}
```
zprobst, line 13:
```suggestion
            {"S": "some_string"}
            {"N": "12345"}
```

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+    if '.' in value:
+        return float(value)
+    return int(value)
+
+DEFAULT_HANDLER = (convert_scalar, leave_untouched)
+
+def convert_value(dynamo_object: Dict[str, Any]):
+    attribute_type, value = DynamoObject(dynamo_object).items()
+    result_handler, type_handler = ATTRIBUTE_TYPE_DESCRIPTOR_HANDLERS.get(attribute_type, DEFAULT_HANDLER)
+    return result_handler(value, type_handler)
```
zprobst, line 51: At first, I was going to complain that this function name isn't particularly communicative. A convert function probably needs to describe the `from` -> `to` of the conversion to be meaningful. 

Then I was looking at it. In reality, this thing has [Inappropriate Intimacy ](https://refactoring.guru/smells/inappropriate-intimacy) with `DynamoObject`. Maybe make this behavior a method on `DynamoObject` called something like `to_python_type` or the like.

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+    "L": (convert_array, convert_value)
+}
+
+class DynamoRowConverter:
+    def __init__(self, item: Dict[str, Any]) -> None:
+        self.item = item
+
+    # { key: {"S": "string"}, secondary_key: { "M": {Stuff} }, tertiary_key: { "L": [Stuff] } }
+    def convert_row(self):
+        return {attribute_name: convert_value(attribute_value) for attribute_name, attribute_value in self.item.items()}
```
zprobst, line 73: Following the suugestion above, I think the need for `convert_value` could go away if you create a factory method for `DynamoRowConverter` called something like `from_raw_row_data` that takes `Dict[str, Any]` and converts the `Any` to a `DynamoObject`. The constructor would then take `Dict[str, DynamoObject`. All of that would lead this to being:

```suggestion
        return {col_name: attr.to_python_type() for col_name, attr in self.column.items()}
```

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
+from logging import getLogger
+from typing import Any, Dict, List, AsyncGenerator
+
+from ...credential_utils import AwsClientFactory
+from ...extractor import Extractor
+
+
+class DynamoObject:
```
zprobst, line 8: I believe Dynamo calls these values `attributes` so perhaps: 

```suggestion
class DynamoAttribute:
```

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+        return {attribute_name: convert_value(attribute_value) for attribute_name, attribute_value in self.item.items()}
+
+
+def verify_list_consistency(value_list: List[Any]):
+    type_set = set([type(value) for value in value_list])
+    if len(type_set > 1):
+        return "L"
+    return REVERSE_ORDER_DESCRIPTOR_HANDLERS[type_set[0]] + "S"
+
+def boolean_to_dynamo_format_converter(value: None | bool):
```
zprobst, line 82: I think this needs an explication on why `None == True` here. Also if `value` is a boolean at all, it returns false which seems strange.

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+    return True if value is None else False
+
+def convert_value_to_dynamo_format(value: Any):
+    value_type, handler = REVERSE_ORDER_DESCRIPTOR_HANDLERS[type(value)]
+    if (value_type == "LIST"):
```
zprobst, line 87:
```suggestion
    if value_type == "LIST":
```

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+    # { key: {"S": "string"}, secondary_key: { "M": {Stuff} }, tertiary_key: { "L": [Stuff] } }
+    def convert_row(self):
+        return {attribute_name: convert_value(attribute_value) for attribute_name, attribute_value in self.item.items()}
+
+
+def verify_list_consistency(value_list: List[Any]):
+    type_set = set([type(value) for value in value_list])
+    if len(type_set > 1):
+        return "L"
+    return REVERSE_ORDER_DESCRIPTOR_HANDLERS[type_set[0]] + "S"
```
zprobst, line 80: This is clever in the generic, but it doesn't work for all inputs. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html

For instance, a list of maps needs to stay as a list of maps because there is no map set. 

I'd check if its part of the 'set' types.

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+    return True if value is None else False
+
+def convert_value_to_dynamo_format(value: Any):
+    value_type, handler = REVERSE_ORDER_DESCRIPTOR_HANDLERS[type(value)]
+    if (value_type == "LIST"):
+        value_type = verify_list_consistency(value)
+    converted_value = handler(value)
+    return DynamoObject({value_type: converted_value})
+
+def recursive_dynamo_format_converter(value: Dict[str, Any] | List[Any]):
```
zprobst, line 92: These appear to only be used in the map below.... Why not just make this two functions? One that handles lists and the other that handles maps. Having the if check here makes things less explicit.

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+    dict: ("M", recursive_dynamo_format_converter),
+    list: ("LIST", recursive_dynamo_format_converter)
+}
+
+
+def convert_value_list_to_dynamo_format(value_list: List[Any]):
+    return_values = []
+    for value in value_list:
+        return_values.append(convert_value_to_dynamo_format(value))
+    return return_values
```
zprobst, line 114:
```suggestion
    return list(map(convert_value_to_dynamo_format, value_list))
```

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class DynamoConditional:
+    @classmethod
+    def from_conditional_data(cls, conditional_data: Dict[str, Any]):
+        return cls(**conditional_data)
+    
+    def __init__(self, AttributeValueList: List[Any], ComparisonOperator: str):
```
zprobst, line 122:
```suggestion
    def __init__(self, attribute_value_list: List[Any], comparison_operator: str):
```

I think the code needs too keep with standard python naming conventions to keep consistent with the rest of the code base.

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+class DynamoDBExtractor(Extractor):
+    @classmethod
+    def from_file_data(
+        cls,
+        table: str,
+        scan_filter: Dict[str, Any] = {},
+        limit: int = 100,
+        **aws_client_args,
+    ):
+        print(aws_client_args)
```
zprobst, line 151:
```suggestion
```

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+        self.logger = getLogger(self.__class__.__name__)
+
+    def scan_table(self):
+        items = []
+        should_continue = True
+        parameters = { "TableName": self.table, "Limit": self.limit}
+        if self.scan_filter:
+            parameters.update({"ScanFilter": self.scan_filter})
+
+        while (should_continue):
```
zprobst, line 180:
```suggestion
        while should_continue:
```

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+        scan_filter: Dict[str, Any],
+        client,
+    ) -> None:
+        self.table = table
+        self.client = client
+        self.scan_filter = DynamoFilterConverter.from_file_data(scan_filter).generate_filter_converter()
+        self.limit = limit
+        self.logger = getLogger(self.__class__.__name__)
+
+    def scan_table(self):
```
zprobst, line 173: Scan table should be a generator here. Otherwise we need to read the whole table into memory before we can start yielding data into the pipeline.

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+
+        while (should_continue):
+            response: dict = self.client.scan(
+                **parameters
+            )
+            items += response["Items"]
+            last_evaluated_key = response.get("LastEvaluatedKey", None)
+            if last_evaluated_key is not None:
+                should_continue = True
+                parameters.update({"ExclusiveStartKey": last_evaluated_key})
```
zprobst, line 188:
```suggestion
                parameters["ExclusiveStartKey"] = last_evaluated_key)
```

No need to allocate a dict here.

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+            else:
+                should_continue = False
+        return items
+
+    async def extract_records(self) -> AsyncGenerator[Any, Any]:
+        records = self.scan_table()
+        converter = DynamoRowConverter
+        for record in records:
+            converted = converter(record).convert_row()
+            print(converted, '\n\n')
```
zprobst, line 198:
```suggestion
```

nodestream/pipeline/extractors/stores/aws/dynamodb_extractor.py
```diff
@@ -0,0 +1,199 @@
...
+                parameters.update({"ExclusiveStartKey": last_evaluated_key})
+            else:
+                should_continue = False
+        return items
+
+    async def extract_records(self) -> AsyncGenerator[Any, Any]:
+        records = self.scan_table()
+        converter = DynamoRowConverter
+        for record in records:
+            converted = converter(record).convert_row()
```
zprobst, line 197:
```suggestion
        for record in self.scan_table():
            converted = DynamoRowConverter(record).convert_row()
```

zprobst (approved):

PR #333, opened 2024-07-09
Interpreter Schema Generation v2

angelosantos4:
In this PR, we rework the way that the interpreter generates a schema based on the underlying interpretations. 

**Changes:**
- Pre-initialization check for schema correctness given `VALIDATE_PIPELINE_SCHEMA` environment variable is set. This schema introspection fix assumes that there is only one source node to be interpreted under any ingest. A check is made for the following:
  - Interpreter: Only one of the before iteration phase or interpretations phase can generate source nodes.
  - MultipleInterpretationPass: If one of the passes contains a source node, all of the other passes must contain a source node.
  - InterpretationPass: Only one of the interpretations can assign a source node.
  - SwitchInterpretations: If one of the branches contains a source node, all of the other passes including the default branch must contain a source node.
- For GraphQL schema generation, LocalDateTime objects are now DateTime objects.

- The before iteration interpretations are now accounted for within the schema introspection. (We acknowledge that either the before iteration and interpretation phases can act as source node creators.)
- Introspection is now broken down into context levels for each component that expands schema from children. For example: 
     1. Each pipeline has their own isolated context level.
     2. For a Multiple Interpretation pass, each interpretation pass has their own context level that is distinct but inherits data from the interpreter (i.e if there is a before iteration phase with relationships).
     3. For a switch interpretation,  each {key: interpretations}, has their own context, ONLY if it is a source node generator. otherwise, each pass is added to the overall interpreter context. 

    
- The switch interpretation now represents the interpretations for each key as an interpretation pass. 
- Source node creators are now added to the end of the schema generation process.

**Additional work/Unsolved Problems:**
- Key matching between switch interpretations within a reachable context. If keys match between switch interpretations at different context levels, we should be able to isolate the context for those relationships. 

For instance: 
```yaml
switch:
  key: A
    - some relationship A
  key: B
    - some Relationship B 
switch:
  key: A 
    - some source node A
  key: B
    - some source node B 
```
In the schema generation, source node A will involve both relationship A and relationship B, even though it is not possible for source node A to have relationship


nodestream/interpreting/interpretations/properties_interpretation.py
```diff
@@ -25,9 +25,11 @@ def interpret(self, context: ProviderContext):
         self.properties.apply_to(context, source.properties, self.norm_args)
 
     def expand_schema(self, coordinator: SchemaExpansionCoordinator):
+        property_list = [property_name for property_name in self.properties]
         coordinator.on_node_schema(
             self.expand_source_node_schema,
             alias=SourceNodeInterpretation.SOURCE_NODE_TYPE_ALIAS,
+            property_list=property_list,
```
zprobst, line 32:
```suggestion
        property_list = [property_name for property_name in self.properties]
        coordinator.on_node_schema(
            self.expand_source_node_schema,
            alias=SourceNodeInterpretation.SOURCE_NODE_TYPE_ALIAS,
            property_list=list(self.properties),
```

nodestream/interpreting/interpretations/switch_interpretation.py
```diff
@@ -28,67 +40,79 @@ class SwitchInterpretation(Interpretation, alias="switch"):
...
         self.normalization = normalization or {}
         self.fail_on_unhandled = fail_on_unhandled
+        self.verify_completeness()
 
     @property
-    def distinct_branches(self) -> bool:
+    def all_interpretations(self) -> Iterable[Interpretation]:
+        yield from list(self.branches.values()) + (
+            [self.default] if self.default is not None else []
+        )
```
zprobst, line 86: This appears clearer to me what is happening. 

```suggestion
        yield from self.branches.values()
        if self.default:
            yield self.default
```

nodestream/project/pipeline_definition.py
```diff
@@ -165,4 +165,5 @@ def initialize_for_introspection(self) -> Pipeline:
         return self.initialize(PipelineInitializationArguments.for_introspection())
 
     def expand_schema(self, coordinator: SchemaExpansionCoordinator):
+        print(self.name)
```
zprobst, line 168: I think we can get rid of this.

zprobst (dismissed):

zprobst (approved):

PR #335, opened 2024-07-17
AWS Credentials Refresh when Not Assuming Roles

zprobst:
By default, if we do not assume a role, we grab a basic boto score session which gets the credentials without refreshing. This PR introduces a change to fall back on a regular boto3 session which will fetch again from the provider chain.

### DIFF
--- a/nodestream/pipeline/extractors/credential_utils.py
+++ b/nodestream/pipeline/extractors/credential_utils.py
@@ -51,13 +51,11 @@ class AwsClientFactory:
         session._credentials = refreshable_credentials
         return session
 
-    def assume_role_if_supplied_and_get_session(self):
-        if self.assume_role_arn:
-            return self.get_boto_session_with_refreshable_credentials()
-        return Session()
-
     def make_client(self, client_name: str):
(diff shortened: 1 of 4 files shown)


nodestream/pipeline/extractors/credential_utils.py
```diff
@@ -51,13 +51,11 @@ def get_boto_session_with_refreshable_credentials(self):
...
+            return boto3.Session(
+                botocore_session=botocore_session, **self.session_args
+            ).client(client_name)
 
-    def make_client(self, client_name: str):
-        session = self.assume_role_if_supplied_and_get_session()
-        return boto3.Session(botocore_session=session, **self.session_args).client(
-            client_name
-        )
+        return boto3.client(client_name, **self.session_args)
```
angelosantos4, line 61: So this base client will refresh credentials using the user's arn by default?

zprobst: Good call. Turns out It will in some cases... but not all: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html 

I've updated it.

angelosantos4 (approved):

PR #338, opened 2024-07-29
Refactor File Handling Code to Fix S3 Incompatability Issues

zprobst:
In this PR, we a complete refactoring to the internals of the file handling logic that allows for increased interoperability and maintainability between file handling logic and sources. To accomplish this, the PR introduces the following changes: 

- Rename `SupportedFileFormat` and `SupportedCompressedFileFormat` to `FileCodec` and `CompressionCodec` respectively to provide a better naming for their changed roles. Now, these types simply provided the decode behaviors and do not own a file in anyway.
- Introduce `ReadableFile` and its subclasses which define methods for building reader types from `self` as well as defining the `path` of the file (which can be a non-real path).
- Introduce `FileSource` and its subclasses which essentially work to build instances of `ReadableFile` and return them to a single unified file extractor behavior.
- Add dramatically better documentation as to what the code does across the entire `files` module. 
- Deprecate the old extractor types for future removal.

### DIFF
--- a/nodestream/pipeline/extractors/stores/aws/s3_extractor.py
+++ b/nodestream/pipeline/extractors/stores/aws/s3_extractor.py
@@ -72,17 +72,13 @@ class S3FileSource(FileSource):
     def find_keys_in_bucket(self) -> Iterable[str]:
         # Returns all keys in the bucket that are not in the archive dir
         # and have the prefix.
-        print("find keys in bucket")
-        print(self.s3_client)
         paginator = self.s3_client.get_paginator("list_objects_v2")
         page_iterator = paginator.paginate(Bucket=self.bucket, Prefix=self.prefix)
         for page in page_iterator:
             keys = (obj["Key"] for obj in page.get("Contents", []))
-            print(keys)
             yield from filter(lambda k: not self.object_is_in_archive(k), keys)
 
     async def get_files(self):
-        print("get files")
         for key in self.find_keys_in_bucket():
             yield S3File(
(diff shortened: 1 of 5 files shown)


tests/unit/pipeline/extractors/test_files.py
```diff
@@ -27,6 +25,17 @@ def fixture_directory():
         yield tempdir
 
 
+@pytest.fixture
+def unspported_file(fixture_directory):
```
angelosantos4, line 29: *supported

nodestream/pipeline/extractors/stores/aws/s3_extractor.py
```diff
@@ -1,91 +1,118 @@
...
+            yield S3File(
+                key=key,
+                s3_client=self.s3_client,
+                bucket=self.bucket,
+                archive_dir=self.archive_dir,
+                object_format=self.object_format,
+            )
+
+
+class S3Extractor(UnifiedFileExtractor):
```
angelosantos4, line 92: Without this class specifying the file source is to be an S3FileSource, how would we do:

UnifiedFileExtractor.from_file_data(**s3FileSourceParams) in the future?

angelosantos4 (approved):

tests/unit/pipeline/extractors/test_files.py
zprobst: Oops. I'll patch that un in a follow up PR.

Issue #339, opened 2024-07-30
[REQUEST] Implement Unified File Extraction


zprobst:
**Is your feature request related to a problem? Please describe.**
Given the work that is landing / landed in #338, we are in the following state: 

- A `UnifiedFileExtractor` that is agnostic to the source of the files and the file reading methodology by introducing `ReadableFile` and `FileSource` concepts. 
- An `S3Extractor`, `FileExtractor`, and `RemoteFileExtractor` that are all subclasses of `UnifiedFileExtractor` that have no behavior changes aside from a specialized `from_file_data` method that consumes arguments.  These are effectively here for backwards compatibility.
- No way to retrieve files from multiple different sources. I.e ingesting from multiple buckets, or a mix of remote and local files (for example) 

**Describe the solution you'd like**
It seems like a win-win from both a user perspective and code maintenance perspective if we were to do the following: 

- Drop `S3Extractor`, `FileExtractor`, and `RemoteFileExtractor` in favor of directly utilizing the `UnifiedFileExtractor` class. 
- Change the `FileSource` concept to be pluggable like other aspects of `nodestream`. 
- Modify the `UnifiedFileExtractor` class to work a little something like this:

```yaml
- implementation: nodestream.pipeline.extractors:UnifiedFileExtractor
  arguments:
      sources: 
         - type: s3
            bucket: my-s3-bucket
            # ... Any other arguments
        - type: local
           globs:
              - foo/bar/baz.json
``` 

This would obviously constitute a breaking change when we drop `S3Extractor`,  `FileExtractor`, and `RemoteFileExtractor`.  However, it does seem that the rest can be implemented without a breaking change which is awesome.

zprobst:
This has been closed via #338

PR #340, opened 2024-07-30
Add declarative interface to `UnifiedFileExtractor`

zprobst:
This implements the design as discussed in #339

### DIFF
--- a/nodestream/pipeline/extractors/files.py
+++ b/nodestream/pipeline/extractors/files.py
@@ -14,6 +14,7 @@ from typing import (
     AsyncContextManager,
     AsyncGenerator,
     AsyncIterator,
+    Dict,
     Iterable,
     List,
     Tuple,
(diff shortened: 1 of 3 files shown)


nodestream/pipeline/extractors/files.py
```diff
@@ -421,14 +435,14 @@ class LocalFileSource(FileSource):
     can be read by the pipeline.
     """
 
-    @staticmethod
-    def from_globs(*globs: str) -> "LocalFileSource":
+    @classmethod
+    def from_file_data(cls, globs: str) -> "LocalFileSource":
```
angelosantos4, line 439: globs should be a List[str]

zprobst: Indeed

angelosantos4 (approved):

PR #343, opened 2024-07-30
Add log message for when a source yields no files

### LINKED ISSUE #308: [REQUEST] Improve FileExtractor error messaging
In a pipeline containing a FileExtractor:
```
- implementation: nodestream.pipeline.extractors:FileExtractor
  arguments:
    globs:
    - data/my-data.json
```

When the file `my-data.json` does not exist, running the pipeline doesn't produce errors.
Should print some error messages when having issues opening a file.

zprobst:
Closes #308


nodestream/pipeline/extractors/files.py
```diff
@@ -162,6 +163,16 @@ def get_files(self) -> AsyncIterator[ReadableFile]:
...
 
+    def describe(self) -> str:
+        """Return a human-readable description of the file source.
+
+        This method should return a human-readable description of the file source.
+        The description should be a string that describes the file source in a
+        way that is understandable to the user. The description should be
+        concise and informative.
+        """
+        return str(self)
```
angelosantos4, line 174: Maybe this can be a raise NotImplementedError, unless it's expected that __repr__ will be implemented for a subclass.

zprobst: The thinking here is that it provides a "safe" default. I don't want the integration interface to be too large.

angelosantos4: Ah that makes sense, I only just noticed that this is just a description of the file, so returning the subclass name by default works out too.

angelosantos4: Oh there is still a missing unit test here that would encapsulate that case.

zprobst: Added!

angelosantos4 (approved):

PR #346, opened 2024-08-01
Squash Migrations

### LINKED ISSUE #299: [REQUEST] Squash Migrations
Django supports the concept of [Squashing Migrations](https://docs.djangoproject.com/en/5.0/topics/migrations/#migration-squashing). In general the use case is that, as you are developing pipelines, you may alter the schema as you are developing and thus, generate several migrations. Instead of checking them all in (which would be fine) its preferred to have one migration that represents the aggregate set of changes that you made.

Imagine something like this:

```
nodestream migrations squash --from migration-name-1 --to migration-name 2 
nodestream migrations squash --from migration-name-1  # leaving off to picks the head migration 
```

Conceptually, I think this means we need to do the following:

- Get the migration graph at each point 
- Generate a new migration from those two points 
- Discard migrations that are in the `to` graph that are not in the `from` graph. 
- Add the new migration to the graph (`from` graph) 

You can just delete all the migrations that you don't want any more and regenerate the migrations. However, that introduces manual effort to understand the migration graph and feels a little hacky. 

**Additional context**
N/A

zprobst:
This feature adds a new `nodestream migrations squash --from x --to y` command. `x` and `y` would be the names of migrations that you want to squash between. If y is left off, then the last migration is specified. 

Closes #299

### DIFF
--- a/nodestream/schema/migrations/operations.py
+++ b/nodestream/schema/migrations/operations.py
@@ -91,6 +91,62 @@ class Operation(LoadsFromYaml, SavesToYaml):
             "arguments": asdict(self),
         }
 
+    def reduce(self, other: "Operation") -> List["Operation"]:
+        """Reduce this operation with another operation.
+
+        This method should be implemented by subclasses to reduce this
+        operation with another operation. The result of this operation should
+        be a list of operations that are equivalent to applying this operation
+        and then the other operation. If the operations cannot be reduced, then
+        this method should return both operations in a list.
+        """
+        return [self] if self == other else [self, other]
+
+    @classmethod
+    def optimize(cls, operations: List["Operation"]) -> List["Operation"]:
+        """Optimize a list of operations.
+
+        This method should be implemented by subclasses to optimize a list of
+        operations. The result of this operation should be a list of operations
+        that are equivalent to applying all of the operations in the input list.
+        """
+        return OperationReducer(operations).reduce()
+
+
+class OperationReducer:
+    """A class that can reduce a list of operations."""
+
+    def __init__(self, operations: List[Operation]):
+        self.operations = operations
+
+    def reduce(self) -> List[Operation]:
+        """Reduce the list of operations.
+
+        Returns:
+            A list of operations that are equivalent to applying all of the
+            operations in the input list.
+        """
+        current = self.operations
+        while True:
+            reduced = self.reduce_once(current)
+            if reduced == current:
+                break
+            current = reduced
+        return current
+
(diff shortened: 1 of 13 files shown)


nodestream/schema/migrations/operations.py
```diff
@@ -179,6 +245,18 @@ def proposed_index_name(self) -> str:
...
+        if isinstance(other, CreateNodeType) and other.name == self.name:
+            return []
+
+        if (
+            isinstance(other, (AddNodeProperty, DropNodeProperty, RenameNodeProperty))
+            and other.node_type == self.name
+        ):
+            return [self]
+
+        return Operation.reduce(self, other)
```
angelosantos4, line 258: If we make a new Operation like AddNodeLabel and we forget to add it into the list of operations to check for, we can have a situation like the following:

operations = [AddNodeLabel, DropNodeType]
resulting_operations = [DropNodeType, AddNodeLabel] = Operation.reduce(DropNodeType (self), AddNodeLabel (other))
Which would cause an error. 

I think this piece should return Operation.reduce(other, self) so that it maintains the original order of operations 
given that a reduction comparison is not given. 

[CreateNodeType, AddNodeProperty, AddNodeLabel, DropNodeType]
iterationA:
[AddNodeProperty, AddNodeLabel, DropNodeType]
Iteration B:
[AddNodeLabel, DropNodeType]
Iteration C:
[DropNodeType, AddNodeLabel]

angelosantos4: Also this prevents infinite loops I noticed Iteration C above would swap indefinately if AddNodeLabel does the same thing with Operation.reduce(AddNodeLabel, DropNodeType) -> [AddNodeLabel, DropNodeType] which is != [DropNodeType, AddNodeLabel]

tests/unit/schema/migrations/test_operations.py
```diff
@@ -303,3 +303,314 @@ def test_rename_node_type_new_name():
...
+    g = AddRelationshipProperty("Likes", "since")
+    h = AddNodeProperty("Team", "mascot")
+    i = DropRelationshipType("Likes")
+    j = CreateNodeType("Team", {"name"}, {})
+    k = DropNodeProperty("Team", "mascot")
+    dl = DropRelationshipProperty("Likes", "since")
+    m = DropNodeType("Team")
+
+    # Intertwined and shuffled operations
+    operations = [a, f, b, g, c, j, h, d, k, i, e, dl, m]
```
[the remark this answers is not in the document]
zprobst: We discussed this. That can't happen.

angelosantos4 (approved):

PR #364, opened 2024-08-30
Gracefully Close Database Connector

sophiadt:
Wanted a mechanism to properly close the database connector and this PR is chaining the pipeline finish event from the Graph Database Writer all the way down to the database connector.


zprobst (changes requested):
Thanks for the PR @sophiadt! 

I think this approach will work for us. A couple requests:
 - Can we  add `finish` to the ABCs for the executor and the ingest strategies? In the case of the executor I think a reasonable default is to simply `pass`? 
 - Is it possible to add a test case or two to cover the added lines?

zprobst:
I've also approved the CI run. I suspect we may see a linter error or two crop up.

sophiadt:
Hi @zprobst!
I've added the abstract methods and tests now 👍

zprobst (approved):
This looks good now! Thanks for the contribution and collaboration!

PR #374, opened 2025-01-07
Unmarshalling dynamo data officially.

angelosantos4:
Using boto3 deserializer method to unmarshall the dynamo records rather than our own version. In order to use the best standards for type standardization.

### DIFF
--- a/tests/unit/pipeline/extractors/stores/aws/test_dynamo_extractor.py
+++ b/tests/unit/pipeline/extractors/stores/aws/test_dynamo_extractor.py
@@ -3,101 +3,37 @@ from unittest.mock import patch
 import pytest
 
 from nodestream.pipeline.extractors.stores.aws.dynamodb_extractor import (
-    ATTRIBUTE_TYPE_DESCRIPTOR_HANDLERS,
-    DynamoData,
     DynamoDBExtractor,
     DynamoRecord,
-    convert_array,
-    convert_map,
-    convert_scalar,
-    format_boolean,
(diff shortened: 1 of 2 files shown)


tests/unit/pipeline/extractors/stores/aws/test_dynamo_extractor.py
```diff
@@ -3,135 +3,52 @@
...
     ({"N": "100"}, 100),
     ({"N": "1.25"}, 1.25),
-    ({"B": "a31n3c3"}, b"a31n3c3"),
-    ({"NULL": "True"}, None),
-    ({"NULL": "False"}, "False"),
-    ({"BOOL": "True"}, True),
-    ({"BOOL": "False"}, False),
-    ({"SS": ["StringA", "StringB"]}, ["StringA", "StringB"]),
-    ({"NS": ["100", "200", "1.0"]}, [100, 200, 1.0]),
-    ({"BS": ["abcdef", "xyz123"]}, [b"abcdef", b"xyz123"]),
```
angelosantos4, line 84: Notably in these tests, the decoder recognizes *S as actual sets rather than lists. For the most part this will be fine given that dynamo likely implements these as sets internally.

angelosantos4: Also the Booleans effectively return as python booleans rather than strings.

zprobst (approved):
Looks good.

PR #383, opened 2025-01-31
Simply DynamoDB Pagination Using Boto3

zprobst:
Cleans up the implementation of DynamoDB to use boto3 for pagination handling.


angelosantos4 (approved):
Should've wen't deeper on the boto library, glad to see they already have these tools like marshall and paginate

Issue #301, opened 2024-04-29
[REQUEST] Extractor Checkpoints 


zprobst:
**Is your feature request related to a problem? Please describe.**

A common pattern in pipelines is to wake up, extract data from a data lake or object store, and then turn it into a graph. However, this can get tricky when the pipeline throws an unhandled exception, is no longer scheduled, or some other change happens that means that it terminates progress. Assuming the user in some way restarts the pipeline, the pipeline will start from the beginning. 

A quick google search of "data pipeline checkpoints" will reveal that this is a common pattern/solution. While nodestream doesn't endeavor to have every bell and whistle, this one seems like an obvious win for long running pipelines. 

**Describe the solution you'd like**

The proposed solution has several components: 

1. Allow users to specify an interval that checkpoints occur 
2. Provide an interface that allows extractors to produce a checkpoint: `async def checkpoint(self)`. The extractor can return anything that is pickle-able. 
3. Provide an interface that allows extractors to resume from a checkpoint: `async def resume_checkpoint(self, checkpoint):` If this method throws an exception, the error will be logged and `extract_records` will be called like normal. 
4. Introduce a pluggable `ObjectStore` api that can be used for more than just checkpoints. #37 would also benefit from this.  Possible implementations are (`null`, `tempfile`, and `s3`)

**Describe alternatives you've considered**
The only other alternative i can think of is for each step to implement that behavior on its own. However, this is troublesome.

PR #384, opened 2025-02-04
Implement Object Storage APIs

zprobst:
This is the first in a series of PRs intended to address #301. NOTE: Target branch is `0.14` - not `main`.

This PR (on its own) introduces only dead code into the library. As described in the issue description, in order to have extractor checkpoints we need to describe storage apis for the contents of those checkpoints. This PR introduces that core behavior and APIs including:

- Namespacing (which will be used to different checkpoints for different pipelines and versions of that pipeline. 
- Signature Verification - to prevent tampering of those checkpoints
- Object storage itself

### DIFF
--- a/nodestream/pipeline/object_storage.py
+++ b/nodestream/pipeline/object_storage.py
@@ -1,0 +1,298 @@
+import base64
+import hmac
+import hashlib
+import pickle
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


angelosantos4 (approved):

zprobst:
Merging this PR since there is appoval. But if you are coming later, feel free to add comments. This feature is still very much in development.

PR #385, opened 2025-02-05
Working Basic Checkpointing

zprobst:
**Note this is stacked on top of [saga/checkpoints/object-store-apis](https://github.com/nodestream-proj/nodestream/tree/saga/checkpoints/object-store-apis)**

This logic wires object storage through the pipeline and into a testing extractor workload. This proves out the workflow and is working. Follow up PRs can include supporting this for more extractors.

### DIFF
--- a/nodestream/pipeline/pipeline.py
+++ b/nodestream/pipeline/pipeline.py
@@ -5,6 +5,7 @@ from typing import Iterable, List, Tuple
 from ..metrics import Metric, Metrics
 from ..schema import ExpandsSchema, ExpandsSchemaFromChildren
 from .channel import StepInput, StepOutput, channel
+from .object_storage import ObjectStore
 from .progress_reporter import PipelineProgressReporter
 from .step import Step, StepContext
 
@@ -62,7 +63,7 @@ class StepExecutor:
                     if not await self.emit_record(record):
                         return
 
-            async for record in self.step.emit_outstanding_records():
+            async for record in self.step.emit_outstanding_records(self.context):
                 if not await self.emit_record(record):
                     return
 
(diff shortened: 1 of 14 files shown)


nodestream/pipeline/pipeline.py
```diff
@@ -129,12 +130,15 @@ class Pipeline(ExpandsSchemaFromChildren):
     and running the steps in the pipeline.
     """
 
-    __slots__ = ("steps", "step_outbox_size")
+    __slots__ = ("steps", "step_outbox_size", "logger", "logger")
```
angelosantos4, line 133: Why double logger here?

nodestream/pipeline/pipeline_file_loader.py
```diff
@@ -51,6 +53,7 @@ class PipelineInitializationArguments:
     on_effective_configuration_resolved: Optional[Callable[[List[Dict]], None]] = None
     extra_steps: Optional[List[Step]] = None
     effecitve_config_values: Optional[ScopeConfig] = None
+    object_store: ObjectStore = field(default_factory=ObjectStore.null)
```
angelosantos4, line 56: Will there eventually be a front-facing yaml read of some kind to declare an object store kind? Would it be in the pipeline level or nodestream level?

angelosantos4: Oh nevermind, I see it's in the same field as the annotations.

zprobst: Yeah so it'll follow the same pattern. Thats for a follow up PR.

nodestream/pipeline/pipeline_file_loader.py
```diff
@@ -149,19 +152,33 @@ def initialize_with_arguments(self, init_args: PipelineInitializationArguments):
...
+            object_store=init_args.object_store,
+        )
 
 
 class PipelineFile:
     def __init__(self, file_path: Path):
         self.file_path = file_path
         self.logger = getLogger(self.__class__.__name__)
 
+    def file_sha_256(self) -> str:
```
angelosantos4, line 167: Curious on what this is doing a bit more, is this leaving behind a pipeline signature to the object store as part of the verification process? If so can it be renamed to something like create_pipeline_signature?

zprobst: No this is to ensure we don't resume checkpoints from effectively different pipelines despite the file being the same.

zprobst: Verification is only through the HMAC process.

angelosantos4 (approved):

PR #395, opened 2025-02-13
Flag to Disable last_ingested_at update

rreddy2:


nodestream/interpreting/interpretations/source_node_interpretation.py
```diff
@@ -133,6 +136,7 @@ def interpret(self, context: ProviderContext):
             self.creation_rule,
             normalized_key,
             normalized_properties,
+            self.update_last_ingested,
```
zprobst, line 139: Its a bit sad to see this parameter be added to here. Taking a peek at `add_source_node`, it returns the node object. What if we took that return value and executed the logic here? Its one thing to have that if statement be duplicated between the interpretations, but its a bit unfortunate to see it split between the interpreting and model packages.

nodestream/interpreting/interpretations/relationship_interpretation.py
```diff
@@ -94,6 +94,7 @@ class RelationshipInterpretation(Interpretation, alias="relationship"):
         "properties_normalization",
         "key_search_algorithm",
         "node_additional_types",
+        "update_last_ingested",
```
zprobst, line 97: The relationship interpretation creates both the relationship and the node.  Note the `node_` and `relationship_` prefixes to a number of the arguments. Do we want to do that here? Its a little unintuitive right now with its current name.  Does `update_last_ingested` prevent the update to the node, relationship, or both?

rreddy2: Hm that makes sense. I guess we'd have to apply it to both in this case, or create two properties: one for each so that we can assign it granularly (even though in the case that we're doing now, both will be false)

nodestream/interpreting/interpretations/source_node_interpretation.py
rreddy2: I can do that. I'll push the change up

zprobst (approved):

PR #406, opened 2025-03-20
Don't Delete Checkpoint When Other Steps Fail

zprobst:


tests/data/test_snapshot_handling_during_errors.py
```diff
@@ -0,0 +1,38 @@
...
+@pytest.mark.asyncio
+async def test_snapshot_handling_during_errors(mocker):
+    extractor = IterableExtractor.range(0, 100000000)
+    kaboom = ExplodesAfter(5000)
+    storage = mocker.Mock(spec=ObjectStore)
+    storage.namespaced.return_value = storage
+    pipeline = Pipeline((extractor, kaboom), 1000, storage)
+
+    def raise_fatal_error(e):
+        # raise e
```
jbristow, line 31: Is this supposed to be commented out?  

If there's a reason for this to stay, it might be helpful to document the reason so overzealous future contributors don't delete it.

jbristow (dismissed):

tests/data/test_snapshot_handling_during_errors.py
zprobst: It was cruft. I removed it.

jbristow (approved):
I don't like the test solution either but I can't see what where the async process is getting terminated early either.

PR #407, opened 2025-03-20
fix: change jmespath single value behavior

jbristow:
Currently jmespath value providers return only the first item in a list.  This causes a lot of confusion, and this PR attempts to address it.

### DIFF
--- a/nodestream/pipeline/value_providers/jmespath_value_provider.py
+++ b/nodestream/pipeline/value_providers/jmespath_value_provider.py
@@ -9,6 +9,7 @@ from yaml import SafeDumper, SafeLoader
 from .context import ProviderContext
 from .value_provider import ValueProvider, ValueProviderException
 
+
 # `QueryStrategy` is here to provide the seam for different optimizations
 # for executing jmespath queries. We can either execute a "fully fledged"
 # jmespath query or we can implement some simple access patterns that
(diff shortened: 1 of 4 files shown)


nodestream/pipeline/value_providers/jmespath_value_provider.py
```diff
@@ -85,7 +85,12 @@ def search(self, context: ProviderContext):
 
     def single_value(self, context: ProviderContext) -> Any:
         try:
-            return next(self.search(context), None)
+            result = list(self.search(context))
+            if not result:
+                return None
+            if len(result) == 1:
+                return result[0]
+            return list(result)
```
angelosantos4, line 93: Conversation with Jon, the code for single value and search should be merged to have direct identification of whether the data is coming in as a list and using that knowledge to determine how to send the data. (directly return the raw_search). Develop something akin of a !pluck provider that takes the first entry when encountering a list as a single value to replace the prior functionality.

cbadke:
I would very much like to +1 the need to have the jmespath provider support setting lists to properties instead of just pulling the first value from the array.

jbristow:
> I would very much like to +1 the need to have the jmespath provider support setting lists to properties instead of just pulling the first value from the array.

Hey! This inspires me to consider thinking about directing my copious free time to finishing this

zprobst:
@cbadke @jbristow  We have a 0.15 branch already with breaking changes. Is this something either of you would be willing to take across the finish line?

PR #409, opened 2025-03-25
allowing lazy loaded arguments as arguments

AGhafaryy:


nodestream/pipeline/extractors/credential_utils.py
```diff
@@ -1,24 +1,32 @@
...
 from time import time
-from typing import Optional
+from typing import Optional, Union
 from uuid import uuid4
 
 import boto3
 import pytz
 from botocore.credentials import RefreshableCredentials
 from botocore.session import Session
+from nodestream.file_io import LazyLoadedArgument
```
zprobst, line 10: I'd prefer not to have file handling logic buried this far down into nodestream code. The `AwsClientFactory` doesn't need to know that some things are lazy loaded and the like. Its just not part of its mandate and leads to a lack of clear seperation of responsibilities. 

An alternative: Lets remove this logic from here and instead focus our attention on  `storage.py`. I'll put some comments there.

nodestream/project/storage.py
```diff
@@ -40,7 +43,7 @@ def describe_yaml_schema():
         from schema import Optional, Schema
 
         return Schema(
-            {"name": str, "type": str, Optional("hmac_key"): str, Optional(str): object}
+            {"name": str, "type": str, Optional("hmac_key"): LazyLoadedArgument, Optional(str): object}
```
zprobst, line 46: I think it should still be _allowed_, just not _recommended_ to place a string value there. Especially if you are just testing or doing something locally.

nodestream/project/storage.py
```diff
@@ -9,14 +9,17 @@ class StoreConfiguration:
     name: str
     storage_type: str
     arguments: Dict[str, Any]
-    hmac_key: Optional[any] = None
+    hmac_key: Optional[Union[LazyLoadedArgument,str]] = None
 
     def initialize(self) -> ObjectStore:
         store = ObjectStore.from_file_arguments(self.storage_type, **self.arguments)
```
zprobst, line 15: Here we can look to what `Target` does. [Here](https://github.com/nodestream-proj/nodestream/blob/acc6c9500d6cc036bbdbea2d0827bbbaaf77f8b1/nodestream/project/target.py#L32), for example. Target leverages the `LazyLoadedArgument#resolve_if_needed` method to shake out the `LazyLoadedArgument` argument instances and give you _raw_ values. Doing so is how we can avoid transporting the file handling logic all the way to the credential utils command. 

```suggestion
        args = LazyLoadedArgument.resolve_if_needed(self.arguments)
        store = ObjectStore.from_file_arguments(self.storage_type, **args)
```

zprobst (changes requested):

zprobst (dismissed):

zprobst (approved):

PR #410, opened 2025-03-28
adding logging and removing only_one arg

AGhafaryy:


nodestream/project/storage.py
```diff
@@ -60,6 +61,8 @@ class StorageConfiguration:
 
     def initialize_by_name(self, name: str) -> ObjectStore:
         if name not in self.storage_configuration_by_name:
+            logger = logging.getLogger(__name__)
+            logger.info(f"Store configuration '{name}' not found.")
```
zprobst, line 65: What are your thoughts on this? 

Instead of placing the log message here, why don't we push that down to the implementation on the null `ObjectStore` - maybe in its constructor. This gives a general warning every time a `null` `ObjectStore` is created. This code is potentially not the only place object stores could be created.

AGhafaryy: i like that! yeah that makes sense. will change it now.

zprobst (approved):

PR #416, opened 2025-05-22
Adding contextual warnings for all FileCodecs failed parsings & checking for object_format within S3 keys if it exists

rreddy2:


nodestream/pipeline/extractors/files.py
```diff
@@ -299,7 +306,16 @@ class LineSeperatedJsonFileFormat(FileCodec, alias=".jsonl"):
     def read_file_from_handle(
         self, reader: TextIOWrapper
     ) -> Iterable[JsonLikeDocument]:
-        return (json.loads(line.strip()) for line in reader.readlines())
+        try:
```
zprobst, line 309: Did you consider moving the try-catch to the caller of read_file_from_handle ? 

I say that because we could do a few things: 
* log the path-like of the file that failed to load
* We can generalize the message 
* The centralized log means that it works even for file types that are plugged in by users.

rreddy2: I thought it would make more sense to put it in the FileFormat object, because it would be useful if we want to reuse the FileFormat object in the future outside of FileExtractor. 
I see what you mean, that it would be much easier to maintain and we can add more context

rreddy2: I will make the change, I think it's easier to maintain for now

zprobst (approved):

PR #420, opened 2025-06-03
Supporting Dynamic Ingestion Metrics Improving the method of metric propagation.

angelosantos4:
# Metrics System Overhaul

## Overview
This PR introduces a significant enhancement to Nodestream's metrics system, enabling dynamic metric creation and improved progress reporting. The changes focus on making metrics more flexible, informative, and efficient while maintaining backward compatibility.

### Dynamic Metrics
- Introduced support for dynamic metric creation, allowing metrics to be instantiated on-demand
- Added type-specific metrics for node ingestion (e.g., `nodes_upserted_by_type_<type>`)
- Implemented relationship type metrics for better visibility into relationship creation (e.g., `relationships_upserted_by_type_<type>`)

### Channel Metrics
- Added channel size metrics to track data flow between pipeline steps
- Implemented channel registration system to associate metrics with specific input/output steps

### Progress Reporting
- Refactored the progress reporter to operate independently of pipeline runtime
- Enhanced progress indicators to better handle metric updates and reporting
- Improved error handling and reporting in progress indicators

### Testing
- Updated test suite to reflect new metric creation patterns
- Added comprehensive tests for dynamic metric creation and registration
- Improved test coverage for progress reporter and indicator interactions

### Metric Creation
```python
# Example of dynamic metric creation for node types
metric = Metric(
    f"nodes_upserted_by_type_{node_type}",
    f"Number of nodes upserted by type {node_type}"
)
```

### Channel Registration
```python
# Channels can now be registered with step names
channel.register_input("step_name")
channel.register_output("step_name")
```

### Progress Reporting
- Progress reporter now runs as an independent task
- Metrics are ticked at regular intervals (default: 1 second)
- Progress indicators receive updates through a callback system

## Impact
These changes provide several benefits:
1. More detailed insights into pipeline performance
2. Better visibility into data flow between steps
3. More efficient metric creation and management
4. Improved progress reporting reliability
5. Enhanced debugging capabilities

## Migration Notes
- Existing metrics continue to work as before
- New dynamic metrics are created automatically when needed


nodestream/pipeline/channel.py
```diff
@@ -55,7 +75,7 @@ async def get(self):
             object: The object that was retrieved from the channel.
         """
         object = await self.queue.get()
-        self.metrics.decrement(Metric.BUFFERED_RECORDS)
+        Metrics.get().decrement(self.metric)
```
angelosantos4, line 78: Debating this portion. When representing the metrics with an index modulo reporting frequency, we will always have non-important information here? I think this gauge inc and dec model is only good for prometheus and I'm debating also making an accumulator version of these metrics.

nodestream/pipeline/step.py
```diff
@@ -57,8 +57,9 @@ def report_error(
         )
         if fatal:
             self.reporter.on_fatal_error(exception)
+            Metrics.get().increment(NodestreamMetricRegistry.FATAL_ERRORS)
```
angelosantos4, line 60: Perfect way to identify REAL Exceptions. Will use this as a universal dashboard indicator.

Makefile
```diff
@@ -29,11 +29,13 @@ venv: poetry.lock
 format: venv
 	poetry run black nodestream tests
 	poetry run isort nodestream tests
+	poetry run ruff check nodestream tests --fix
 
 .PHONY: lint
 lint: venv
 	poetry run black nodestream tests --check
 	poetry run ruff check nodestream tests
+	poetry run isort nodestream tests --check-only
```
angelosantos4, line 38: I like to imagine one checks and the other fixes. We need these to match the fixes required by the github actions pipeline. (It's always missing.)

nodestream/pipeline/channel.py
```diff
@@ -37,12 +37,32 @@ class Channel:
     prevent one step from overwhelming another step with too many records.
     """
 
-    __slots__ = ("queue", "input_dropped", "metrics")
+    __slots__ = ("queue", "input_dropped", "_metric", "input_name", "output_name")
```
zprobst, line 40: I'd recommend the following change here: 

- Store a `Metric` on the `channel` instead of taking an input and output name. 
- Create a factory method on `Channel` that computes the metric name. 

This has a couple advantages:
- Means we don't need to have this lazy computing metric
- Allows the channel to be reused in different contexts. This implementation assumes that the in and out of the channel can be named.

nodestream/pipeline/channel.py
zprobst: You can implement an accumulator with `inc` and `dec`. Thats fundamentally what is happening with nodes  upserted and the like.  The handler is in charge of how to respond a change in the metrics. All we need to do in this code is capture the metric change. Consider it "emitting an event" more than having a direct impact on the reporting.

nodestream/metrics.py
```diff
@@ -188,39 +223,55 @@ class ConsoleMetricHandler(MetricHandler):
     """A metric handler that prints metrics to the console."""
 
     def __init__(self, command: Command):
-        self.metrics = {x: 0 for x in Metric}
+        self.metrics: dict[Metric, Number] = {
```
zprobst, line 226: Why initialize this for all metrics if we already do it lazily with the `.get(` in `increment` and `decrement`)?

nodestream/metrics.py
```diff
@@ -59,25 +69,66 @@ def increment(self, _: Metric, __: Number):
...
+        Returns a dictionary mapping metric names to their Metric instances,
+        collected from all MetricRegistry subclasses.
+        """
+        all_metrics = {}
+        for subclass in cls._subclasses:
+            # Collect metrics from class variables
+            for attr_name, attr_value in subclass.__dict__.items():
+                if isinstance(attr_value, Metric):
+                    all_metrics[attr_name] = attr_value
+        return all_metrics
```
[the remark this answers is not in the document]
zprobst: I get the idea of distinguishing dynamically created metrics from static ones. But can we annotate that with `Metric` instead of creating this registry concept? It seems very heavy for what we are trying to achieve. What are your thoughts?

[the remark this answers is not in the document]
angelosantos4: So would this be like, find all instances of metric that are statically defined through some kind of Metric.obtain_instances() ? 

I thought this was nice as it included plugins pretty well. So if your custom plugin has metrics it can be registered through this method as well. It is honestly fair that the only `real` use of this distinction is for metric ordering to be defined which only matters in the console output/ json too kind of. 

The concept can also be removed entirely in favor of Metrics.get().increment(metrics). (All handlers dynamically handle metric initialization. Which I can make part of the abstract. )

nodestream/pipeline/channel.py
angelosantos4: Yeah I guess the internal debate was between having one or two metrics per channel. accumulator_metric = Metric("accumulate") and gauge_metric = Metric("gauge"), one increments and the other inc and dec accordingly.

PR #422, opened 2025-06-09
S3 source filtering

jbristow:
Changes the s3 source to download all files matching a prefix and treat them all as if they had the extensions denoted in `S3FileSource.object_format`

### DIFF
--- a/tests/unit/pipeline/extractors/test_files_s3.py
+++ b/tests/unit/pipeline/extractors/test_files_s3.py
@@ -1,7 +1,6 @@
 import bz2
 import gzip
 import json
-import logging
 from typing import Callable
 from unittest.mock import MagicMock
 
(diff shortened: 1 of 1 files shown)


nodestream/pipeline/extractors/files.py
```diff
@@ -391,6 +393,21 @@ def read_file_from_handle(
...
+class TempFile(LocalFile):
+    def __init__(self, path: Path, pointer=None, original_path: str = ""):
+        super().__init__(path, pointer)
+        self.original_path = original_path or self.path.name
+
+    def as_reader(self, cls: type[IOBase]) -> AsyncContextManager[IOBase]:
+        return super().as_reader(cls)
+
+    def path_like(self) -> Path:
+        return super().path_like()
```
zprobst, line 405: This seems redundant. Can it be removed?

nodestream/pipeline/extractors/files.py
```diff
@@ -391,6 +393,21 @@ def read_file_from_handle(
         return [safe_load(reader)]
 
 
+class TempFile(LocalFile):
+    def __init__(self, path: Path, pointer=None, original_path: str = ""):
```
zprobst, line 397: Curious why we represent `original_path` as `str`?

nodestream/pipeline/extractors/files.py
jbristow: it is as you say! 🫡

PR #426, opened 2025-06-27
Splunk extractor

AGhafaryy:


nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+                            "dispatchState" in elem.tag
+                            or elem.get("name") == "dispatchState"
+                        ):
+                            dispatch_state = elem.text
+                            break
+                except Exception as e:
+                    self.logger.warning(
+                        "Failed to parse job status",
+                        extra={"error": str(e), "search_id": search_id},
+                    )
```
angelosantos4, line 249: I see there is a pattern here between parsing things as Json and parsing them as XML. I would recommend creating a parser class that will take an arbitrary object of one of the two types and get the corresponding object a bit more declaratively. 

``` 
class SplunkResponseParser:
    def parse(object: Json | XML )
        ...
```

nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+
+            dispatch_state = "UNKNOWN"
+
+            try:
+                job_status = response.json()
+                dispatch_state = (
+                    job_status.get("entry", [{}])[0]
+                    .get("content", {})
+                    .get("dispatchState")
+                )
```
angelosantos4, line 231: I think there are heavy assumptions here about the state of the JSON, do we know why it is the first field in the entry list? I fear that a parallel job run from a different splunk extractor will make a job, and it will end up being the top of the list  here. Or is it the case that a search can have multiple Jobs? Why is it the first one that we choose that we care about.

nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+            )
+
+            if dispatch_state == "DONE":
+                self.logger.info("Search job completed", extra={"search_id": search_id})
+                return
+            elif dispatch_state == "FAILED":
+                raise RuntimeError(f"Search job failed: {search_id}")
+
+            await asyncio.sleep(2)  # Wait 2 seconds before checking again
+            wait_count += 2
```
angelosantos4, line 267: I think it is more standard to do something akin to:
```
attempts = 0
while should_continue:
    should_continue = success_condition AND attempts < MAX_ATTEMPTS
    attempts += 1
```

nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+                },
+            )
+
+            if dispatch_state == "DONE":
+                self.logger.info("Search job completed", extra={"search_id": search_id})
+                return
+            elif dispatch_state == "FAILED":
+                raise RuntimeError(f"Search job failed: {search_id}")
+
+            await asyncio.sleep(2)  # Wait 2 seconds before checking again
```
angelosantos4, line 266: You can also replace this with a global variable SPLUNK_STATUS_CHECK_PERIOD_SECONDS.

nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+            )
+
+        self.logger.info(
+            "Created search job successfully", extra={"search_id": search_id}
+        )
+        return search_id
+
+    async def _wait_for_job_completion(
+        self, client: AsyncClient, search_id: str, max_wait_seconds: int = 300
+    ):
```
angelosantos4, line 204: Try typing the return types. Instead of nesting try: catches, you can return a completion state of True and False in order to hand off the expected state to the over-arching function.

nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+        password: Optional[str] = None,
+        earliest_time: str = "-24h",
+        latest_time: str = "now",
+        verify_ssl: bool = True,
+        request_timeout_seconds: int = 300,
+        max_count: int = 10000,
+        app: str = "search",
+        user: Optional[str] = None,
+        chunk_size: int = 1000,
+    ) -> "SplunkExtractor":
```
angelosantos4, line 42: Try to seperate the functionality into a Client and have the extract_records

nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+
+    async def _get_job_results(
+        self, client: AsyncClient, search_id: str
+    ) -> AsyncGenerator[Dict[str, Any], None]:
+        """Get results from a completed search job with pagination."""
+        while not self.is_done:
+            params = {
+                "output_mode": "json",
+                "count": str(self.chunk_size),
+                "offset": str(self.offset),
```
angelosantos4, line 281: Why are these strings, do we need to change the input parameters to just intake strigified integers?

nodestream/pipeline/extractors/stores/splunk_extractor.py
angelosantos4: Maybe:

```
for entry in job_status.get("entry", []):
    Do checking on all entries that some condition is true. Or that one of the entries is the one we are looking for.
```

nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+                headers=self._headers,
+                auth=self._auth,
+            )
+
+            if response.status_code != 200:
+                raise HTTPStatusError(
+                    f"Failed to get job results: {response.status_code}",
+                    request=response.request,
+                    response=response,
+                )
```
angelosantos4, line 296: What if the error is a 504, it could be the case that we hit the endpoint often and they would try to rate-limit us.

nodestream/pipeline/extractors/stores/splunk_extractor.py
jbristow: +1 to the interface idea.

nodestream/pipeline/extractors/stores/splunk_extractor.py
jbristow: Also, I feel like this indicates there might be a way to make a more pluggable "client" idea once we figure out a unified retry logic.  🤔

nodestream/pipeline/extractors/stores/splunk_extractor.py
```diff
@@ -0,0 +1,400 @@
...
+            raise RuntimeError(
+                f"Failed to extract search ID from response: {response.text[:500]}"
+            )
+
+        self.logger.info(
+            "Created search job successfully", extra={"search_id": search_id}
+        )
+        return search_id
+
+    async def _wait_for_job_completion(
```
jbristow, line 202: My IDE flags this as having too much "cognitive complexity" (Aka too many nested branches and loops)

It would be good to refactor chunks of this into smaller functions to reduce the amount of brainpower required to figure out what's going on.

tests/unit/pipeline/extractors/stores/test_splunk_extractor.py
```diff
@@ -0,0 +1,657 @@
...
+    )
+
+    assert_that(extractor.verify_ssl, equal_to(True))
+    assert_that(extractor.request_timeout_seconds, equal_to(300))
+    assert_that(extractor.chunk_size, equal_to(1000))
+
+
+# Helper property tests
+def test_splunk_extractor_auth_property_with_token(splunk_extractor):
+    assert_that(splunk_extractor._auth, equal_to(None))  # Token goes in header
```
jbristow, line 105: in general I would avoid testing internal/private properties and functions to avoid making tests that rely on implementation details.

These tests can be handled by `responses` and `pytest-httpx` by having these mock libraries watching for expected auth headers.

tests/unit/pipeline/extractors/stores/test_splunk_extractor.py
```diff
@@ -0,0 +1,657 @@
...
+    assert_that(
+        results[0],
+        has_entries(
+            {
+                "_time": "2023-01-01T10:00:00",
+                "host": "server1",
+                "message": "Login successful",
+            }
+        ),
+    )
```
jbristow, line 405: why not assert the whole result?

```suggestion
    results == [
      {
           "_time": "2023-01-01T10:00:00",
            "host": "server1",
            "message": "Login successful",
      }, 
      {...}, 
      {...}
    ]
```

PR #435, opened 2025-08-26
Implement Record Linage Tracking and Finalization Callback

zprobst:
This PR focuses on refactoring the pipeline code in order to better accommodate and implement a record lineage tracking system. The goal of which is to enable steps, often extractors, to free up resources associated with a yielded record to acknowledge back to systems that a record was truly and completely processed. 

## Refactor 

The current pipeline code is tenuous and hard to introduce change into because of the fact that it is implemented as a series of procedures and not well encapsulated. In these situations, we need to make some cross cutting changes to refactor the pipeline code. 

The refactor itself works on the observation that the previous code was implemented as a procedure so that there could be careful transitions from one state that a step is in to another. In order to remove the procedures, the pipeline was refactored to operate a state machine for each step in the pipeline as well as the output. 

Steps Transition through a state flow that looks like this: 

 `StartStepState` →  `ProcessRecordsState` →  `EmitOutstandingRecordsState` →  `StopStepExecution`

and the pipeline output progresses via a state flow like this:

`PipelineOutputStartState` →  `PipelineOutputProcessRecordsState` →  `PipelineOutputStopState`

This means that both steps and  the pipeline output use the same executor pattern, simplifying the overall architecture.

## Lineage Tracking 

The highlight of this PR i record lineage tracking. This, in short, builds a tree of intermediary and final records produced from every single output record. In other words, we track parent-child relationships of every record in and out of every step. With this information, we are able to know when `nodestream` is 'done' processing a record  and trigger a callback to the originating step when its appropriate. All steps can use this following the same pattern. As an example an extractor is provided below: 

```python
from nodestream.pipeline import Extractor


class TracksLineageWithTheOriginalRecord(Extractor):
    async def finalize_record(self, record):
      print("The following record was returned to me", record)

    async def extract_records(self):
      for i in range(10000):
          yield i 


class TracksLineageWithSomeToken(Extractor):
   def __init__(self, items):
      self.items = items

   async def finalize_record(self, token):
      print("The following record was returend to me", self.items[token])

   async def extract_record(self):
      for index, item in enumerate(self.items):
         yield item, index # yielding a tuple here, the second item here is what is sent back to `finalize_record`
```

## Bug Fixes 

As miscellanies, this PR fixes to minor issues with exception handling in the pipeline logic as well. 

* Fixes a race condition where If an exception is thrown before the  `on_start` callback is triggered, the CLI may not have started the progress spinner that reports error messages. This leads to a separate error on the crash that obscures the original error. The fix is to ensure that the `on_start` is explicitly executed first before actual pipeline processing begins. 
* Fixes a error where it is impossible for the CLI to throw an exception in order to emit a status code other than zero. This is because the hook provided to the CLI to do this (`on_finish`) had a try catch block around it. That try catch was not required because, by the time the pipeline executes this code, there is nothing else to do so there is no reason to swallow the exception in order to protect the integrity of the pipeline. the fix was to simply remove that try catch. 

Due to this behavior, we'll likely need to make a 0.15 release.


nodestream/pipeline/pipeline.py
```diff
@@ -10,122 +11,354 @@
...
+    ):
+        """Create a record from a step's emission of data.
+
+        The `emission` can either be a single value or a tuple of two values.
+        If it is a single value, then it is assumed to be the data for the
+        record. If it is a tuple of two values, then the first value is
+        assumed to be the data for the record and the second value is assumed
+        to be the callback token for the record. If any other value is
+        provided, the data and callback token are both set to the value
+        provided.
```
cbadke, line 47: This is how i figured it would work. Documentation will have to clearly state that after this change `tuple` is not an acceptable record type unless the steps provides some kind of callback token. That is to say, `tuple` record types will only work if the emission type is `tuple[tuple, Any]`.

nodestream/pipeline/pipeline.py
```diff
@@ -10,122 +12,417 @@
...
+        # is simply if our children are useful.
+        self.child_record_count -= 1
+        if self.child_record_count == 0:
+            await self.drop()
+
+    async def drop(self):
+        # If we are being told to drop, then we need to run our callback so
+        # that the step that created us can clean up any resources it has
+        # allocated for this record.
+        await self.originating_step.finalize_record(self.callback_token)
```
cbadke, line 79: should the call to finalize_record be skipped if `data == callback_token`? From lines 58-60, I figured that was the intent. Or is `finalize()` always called and the `callback_token` is just a way for the step author to have the callback call with different data than the original data record?

nodestream/pipeline/pipeline.py
```diff
@@ -10,122 +12,417 @@
...
+    to this state, it will not transition to any other state and end execution.
+    """
+
+    async def execute_until_state_change(self) -> Optional[ExecutionState]:
+        self.call_ignoring_errors(self.reporter.on_finish_callback, self.metrics)
+        return None
+
 
-        self.call_handling_errors(self.reporter.on_finish_callback, metrics)
+class Exectutor:
```
cbadke, line 401: should be `Executor` 😄

nodestream/pipeline/pipeline.py
zprobst: Agreed...

nodestream/pipeline/pipeline.py
cbadke: I can see the argument and temptation to always call `finalize_record` but to change what is passed.

The concern I have with calling finalize in either case is the inconsistency and element of surprise. If I return one thing from `process_record` I get one behaviour, if I return a magic tuple, I get a different behaviour.

My feeling is that there should generally be one way for things to work. If you only call `finalize` if they return `tuple` with a callback object, the system is explicitly forcing the developer to say "let me know when this is finished processing". If they want the original message, they can return `(record, record)`.

Maybe it should even be a flaggable feature but that would complicate matters I imagine.

cbadke: No matter which way you choose, the `tuple` vs `not tuple` behaviour is going to confuse someone at some point. :-/

zprobst: I agree with your thought process here. I've solicited @jbristow to get another perspective. I definitely see this both ways.

jbristow: Thinking about this as if I were writing Haskell or F# code, it feels like there should be "current-context" struct being passed from step to step that contains the result rather than a result.

That way you could bind things to the context that would be called by specific lifecycle events.  Maybe make some Global hooks like "after-successful-ingest" or whatever.

I apologize for bringing up monad adjacent thought while discussing Python code.

cbadke: I had a similar thought early on but wasn't sure on the appetite to change the interface. I think it could work.

1. I'm not sure we should be singularly focused on Extractors for this feature, there might be cases for other constructs (like Transforms?) that could want to access this mechanism once it's available.
2. It could feel strange to have components return this record object but receive just the `data` element from the upstream step. Should all steps receive this Record wrapper instead? I would guess the interface would use the base class.

nodestream/pipeline/step.py
```diff
@@ -110,6 +106,8 @@ class Step:
     asynchronous context. They can process records and emit new records.
     """
 
+    tracks_lineage: bool = False
+
```
cbadke: Would there be any developer benefit to encapsulating this as a new subclass of `Step`? Then `finalize_record` could only be declared on that interface?

I feel like this could avoid confusion for Step developers that don't need this finalizing behaviour since if that boolean is false then `finalize_record` is never called.

```python
def FinalizingStep(Step):
    async def finalize_record(self, callback_token: object):
         """Finalize a record.
        This method is called when a record produced by this step has been
        fully processed by all downstream steps. It is not called for records
        that are not produced by this step.
        """
        pass
```

nodestream/pipeline/pipeline.py
```diff
@@ -1,131 +1,436 @@
...
-    The `StepExecutor` is responsible for starting, stopping, and running a
-    step in a pipeline. It is used to execute a step by passing records
-    between the input and output channels of the step.
-    """
 
-    __slots__ = ("step", "input", "output", "context")
+@dataclass(slots=True)
+class Record:
+    """A `Record` is a unit of data that is processed by a pipeline."""
+
```
cbadke, line 22: This is only a framework facing class so maybe not critical but I wonder if naming this `Record` will cause some confusion. As I'm reading through the code, I'm realizing that throughout the system `record` is used to refer to the input and output of Steps. But now we have a new `Record` class where that object that is often called a record is the `data` property of this class. Could get confusing for folks not deeply entrenched in the framework?

nodestream/pipeline/pipeline.py
```diff
@@ -1,131 +1,436 @@
...
+            emission (Any): The emission from the step.
+            originated_from (Optional[Record], optional): The record that
+                this record was emitted from. Defaults to None.
+
+        Returns:
+            Record: The record created from the emission.
+        """
+        data = callback_token = emission
+        if isinstance(emission, tuple) and step.tracks_lineage:
+            data, callback_token = emission
```
cbadke: Would it be beneficial for `callback_token == None` when `steps.tracks_lineage == None`? Would make it explicitly clear than no token was actually communicated out from the step.

nodestream/pipeline/pipeline.py
```diff
@@ -177,14 +481,20 @@ async def run(self, reporter: PipelineProgressReporter):
...
+        # position in the list that makes its way to the gather call. It will
+        # break the pipeline because we need to ensure the on_start_callback
+        # is called before any operations occur in the actual steps in the
+        # pipeline. To do this, we need to make sure that the first coroutine
+        # scheduled is the one that calls the on_start_callback.
         current_input, current_output = channel(
             self.step_outbox_size, current_output_name, current_input_name
         )
-        pipeline_output = PipelineOutput(current_input, reporter)
+        executors.append(Executor.pipeline_output(current_input, reporter))
```
cbadke, line 497: I might be misremembering but I thought we needed the `reporter` to be _first_ in the list in order to fix the race?

tests/unit/pipeline/test_record.py
```diff
@@ -0,0 +1,254 @@
...
+    """Test Record.from_step_emission with tuple (data, token)."""
+    step = Mock(spec=Step)
+    data = {"test": "data"}
+    token = "callback_token"
+
+    record = Record.from_step_emission(step, (data, token))
+
+    assert record.data == data
+    assert record.callback_token == token
+    assert record.originating_step == step
```
cbadke: How does this test pass if `tracks_lineage` isn't set on the `Step`? This seems like an issue with a number of tests in this file?

Maybe Mock defaults bools to `True`? If so, I feel like it would be clearer to be explicit about that value.

tests/unit/pipeline/test_step.py
```diff
@@ -102,3 +102,31 @@ async def test_step_context_report_warning_message(mocker):
...
+
+@pytest.mark.asyncio
+async def test_default_finalize_record_does_nothing(mocker):
+    step = Step()
+    token = "test_token"
+    await step.finalize_record(token)
+    # Should not raise any exceptions and should do nothing
+
+
+@pytest.mark.asyncio
```
cbadke: If I'm not mistaken, I think this test just tests that You can call a function. The `mock_finalize` is defined, attached to the `step` object and then called. I'm not sure this is testing anything.

tests/unit/pipeline/test_step_finalize_record.py
```diff
@@ -0,0 +1,270 @@
...
+
+@pytest.mark.asyncio
+async def test_step_finalize_record_default_behavior():
+    """Test that default finalize_record implementation does nothing."""
+    step = Step()
+
+    # Should not raise any exceptions
+    await step.finalize_record("any_token")
+    await step.finalize_record(None)
+    await step.finalize_record({"complex": "token"})
```
cbadke: I think this is redundant to `test_default_finalize_record_does_nothing` in `test_step.py`?

tests/unit/pipeline/test_step_finalize_record.py
```diff
@@ -0,0 +1,270 @@
...
+    # Drop second child
+    await child2.drop()
+
+    # Now both children and parent should be finalized
+    assert child_step.finalize_calls == ["child1_token", "child2_token"]
+    assert parent_step.finalize_calls == ["parent_token"]
+
+
+@pytest.mark.asyncio
+async def test_finalize_record_async_behavior():
```
cbadke: Is this actually testing that `Record.drop()` calls the finalize with `await`?

cbadke (commented):
I'm really excited to see how all this is coming together. Just a few notes and questions.

nodestream/pipeline/step.py
zprobst: Interesting idea... thought about it a bit. Right now we have class hierarchies that look like this: 

```mermaid
graph LR
    A[Step] --> B(Transformer)
    B --> C[MyAwesomeTransformer]
```

Lets assume that we want to add finalization to our `MyAwesomeTransformer` by inheriting from `FinalizingStep`. We'd need to have a class hierarchy like this: 

```mermaid
graph LR
    A[Step] --> B(Transformer)
    B --> C[MyAwesomeTransformer]
    D[FinalizingStep] --> C
    A --> D
```

This creates a... confusing class hierarchy and can lead to weird to weird [MRO](https://medium.com/@ruitcatarino/understanding-pythons-method-resolution-order-mro-f7cbcec36993) issues.

Then imagine we have a `ApronSpringsStep` that gets notified every time we have operate on a give record.

```mermaid
graph LR
    A[Step] --> B(Transformer)
    B --> C[MyAwesomeTransformer]
    D[FinalizingStep] --> C
    A --> D
    E[ApronSpringsStep] --> C
    A --> E
```

This violates my personal rule for relatively shallow, flat hierarchies of classes. The more cases we add to this example the more it feels that its really the same case with the implementer of `Step` choosing to do something or not depending on the cases.

nodestream/pipeline/pipeline.py
zprobst: Thats a good point... I'll workshop a better name

nodestream/pipeline/pipeline.py
zprobst: I think it is? Line `484` we create a blank list and `497` is the first we append to it so it will have position 0. Do you think its better if we create it there to be a touch more clear about that. Something like:

```python
executors = [Executor.pipeline_output(current_input, reporter)]
```

Does that seem more appropriate?

nodestream/pipeline/step.py
cbadke: Yep, that's a really good point. As I look at this, it feels like `Finalizing` is more of a [Protocol](https://typing.python.org/en/latest/spec/protocol.html) than a SubClass. Would that feel any better?

It's a bit of an abuse because utilizing the protocol changes the frameworks treatment of `Step` outputs so maybe it's still not a great idea. I think I'm just trying to address the bad feeling of a boolean behavior flag and a method that is unimportant to most use cases.

I leave it to your judgement on how you want that interface and experience to work.

nodestream/pipeline/pipeline.py
cbadke: I see. I was completely misinterpreting this section. I think I'm not fulling grokking all of the `Executor` abstractions and channel management and that was clouding my read of it.

nodestream/pipeline/pipeline.py
zprobst: I've gone with `RecordContext`

nodestream/pipeline/pipeline.py
zprobst: My pendulum has swung the other way on this. I think always calling it and calling and not having a flag is the most predictable pattern.

nodestream/pipeline/pipeline.py
zprobst: All good... this PR is doing a lot in it.

tests/unit/pipeline/test_step.py
zprobst: Probably teh tests are pretty bad. I'll go through them

tests/unit/pipeline/test_step_finalize_record.py
zprobst: Yes it is. Will remove. Ai tests .... smh

nodestream/pipeline/step.py
zprobst: I think I've come down into there isn't really a need to distinguish a protocol  or subclass to avoid the bools. Not having it in this case is the same as doing nothing. For `Step` is is always a reasonable default implementation that we can rely on. This case its just `pass`.

zprobst:
I think due to the complexity and nuance in this change, we are due for merging this into a new breaking release. I see this as a good opportunity to make some other small changes we need to make a-la #407

nodestream/pipeline/pipeline.py
cbadke: Make sure to update docs and advertise the breaking change that any tuple returned from a step will have the last element stripped off.

nodestream/pipeline/pipeline.py
```diff
@@ -1,131 +1,436 @@
...
+    pass
 
-    The `StepExecutor` is responsible for starting, stopping, and running a
-    step in a pipeline. It is used to execute a step by passing records
-    between the input and output channels of the step.
-    """
 
-    __slots__ = ("step", "input", "output", "context")
+@dataclass(slots=True)
+class RecordContext:
```
ccloes, line 20: Did you consider having a "max-depth" param to limit the amount of Records in flight?  Or is this handled natively elsewhere?  Could a long running job with lots of records cause memory consumption issues?

nodestream/pipeline/pipeline.py
```diff
@@ -1,131 +1,436 @@
...
+        self.child_record_count -= 1
+        if self.child_record_count == 0:
+            await self.drop()
+
+    async def drop(self):
+        # If we are being told to drop, then we need to run our callback so
+        # that the step that created us can clean up any resources it has
+        # allocated for this record if it opts into the feature
+        # (by implementing the `finalize_record` method).
+        await self.originating_step.finalize_record(self.callback_token)
```
ccloes, line 76: Is it possible for finalize_record to fail here?  And if so, is it a choice to not handle errors here and allow them to propogate up?

ccloes (approved):

nodestream/pipeline/pipeline.py
zprobst: This is effectively controlled with the `--step-outbox-size` parameter that controls the amount of in-flight records.

nodestream/pipeline/pipeline.py
zprobst: Really good question... it _is_ possible given that it is a `Step`, which can do anything, we are calling here. 

It was a decision. Generally all calls to steps are allowed to throw exceptions given the contract that its thrown for something truly exceptional that _should_ halt the pipeline in due time. Nodestream cannot really know whether something is retryable or a unrecoverable issue. So if you, as a step, throw an exception, nodestream catches and handles them during execution. In this case, throwing an exception is caught up the chain in the part of the pipeline code that manages the execution of the step and gently ends the pipeline.

PR #439, opened 2025-10-08
Allowing additional params to be sent to KafkaConsumerFactory

rreddy2:


nodestream/pipeline/extractors/streams/kafka.py
```diff
@@ -17,12 +17,14 @@ def build_consumer_config(
         security_protocol: str,
         group_id: str,
         bootstrap_servers: List[str],
+        **extra_params,
```
jbristow: Why would this not be:
```suggestion
        **kwargs: Any,
```

jbristow (dismissed):

nodestream/pipeline/extractors/streams/kafka.py
rreddy2: I'm okay with keeping it consistent. I was basing my code changes off of the changes from etwpipeline

jbristow (approved):

PR #445, opened 2025-11-21
Attempting to add schema provenance for node and relationship.

ccloes:
Adding the ability to introspect nodes and relationships.


ccloes:
This is still a work in progress...

PR #446, opened 2025-12-15
Updating the schema expansion to include additional labels.

angelosantos4:
Schema Generation Change: 
- Additional labels within an ingest will now expose all the relationships and properties available to the original node_type or related_node_type.

For example: 

```
      - - type: source_node
          node_type: Player
          key:
            player_id: !jmespath player_id
          properties:
            name: !jmespath name
            overall: !jmespath overall
          additional_types:
            - Person
        - type: relationship
          node_type: Country
          relationship_type: ORIGINATES_FROM
          node_key:
            name: !jmespath nationality
```

Will expand into a Player -ORIGINATES_FROM> Country and Person - ORIGINATES_FROM > Country
Player and Person will also share and expose the same properties and keys when the schema is expanded. 

This has the side effect of creating indexes for additional labels with the same keys, and checks for key consistency across additional labels.

### DIFF
--- a/nodestream/interpreting/interpretations/relationship_interpretation.py
+++ b/nodestream/interpreting/interpretations/relationship_interpretation.py
@@ -227,6 +227,19 @@ class RelationshipInterpretation(Interpretation, alias="relationship"):
                 node_type=self.node_type.value,
             )
 
+            # Expand schemas for node additional types with the same structure
+            for additional_type in self.node_additional_types:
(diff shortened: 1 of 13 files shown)


zprobst (approved):
I've merged this onto the 0.15 branch as the feature at its core works but will likely break things. We can append some mitigations or options for it on the 0.15 branch

PR #451, opened 2026-02-23
Copy pipeline parallelization.

angelosantos4:
### Copy Pipeline Overhaul

- **Concurrent copier** — `Copier.create()` factory accepts a `concurrency_limit`; any value > 1 automatically builds a `ConcurrentCopier` that fetches multiple node/relationship types in parallel, bounded by a semaphore. Nodes and relationships are processed in separate phases so they never starve each other.
- **Type histogram & priority sorting** — Before copying, the copier previews counts for every node and relationship type via `TypeRetriever.preview_node_count` / `preview_relationship_count`, logs a descending histogram, and processes the largest types first.
- **New CLI tunables for copy** — Added `--concurrency-limit`, `--batch-size`, `--step-outbox-size`, `--flush-concurrency`, `--connector-option key=value`, and `--reporting-frequency` options to the `copy` command.
- **Progress reporting** — The `copy` command now wires up the same progress reporter (spinner / JSON) used by `run`, including elapsed time and records-per-second display.
- **Schema without additional types** — Copy now builds the schema with `include_additional_types=False` so only concrete pipeline-declared types are copied. This limits the amount of duplicate work done by re-ingesting nodes under a different label alias.

### Writer Concurrency

- **`ConcurrentGraphDatabaseWriter`** — New subclass that rotates ingest strategies and flushes completed batches in background tasks. A semaphore (`max_flush_lanes`) provides back-pressure. The base `GraphDatabaseWriter` remains unchanged; `flush_concurrency > 1` in the factory methods automatically returns the concurrent variant.
- **Target writer overrides** — `Target.make_writer()` now accepts `connector_overrides` and `**writer_overrides`, letting callers (e.g. the copy command) customise batch size, flush concurrency, and connector settings at runtime.

### Metrics Improvements

- **`set_value` API** — Added `Metric.set_value` / `Metrics.set_value` across all metric handlers (Prometheus, Console, JSON, Null, Aggregate) for gauge-style metrics that should not accumulate.
- **New gauges** — `writer_pending_records`, `writer_active_flush_lanes`, `orchestrator_queue`, and `active_queries` provide real-time observability into the copy and writer pipelines.
- **Channel buffer metric is now a gauge** — The `buffered_*` channel metric switched from `accumulate=True` to `accumulate=False` so it reflects the current queue depth, not a running total.
- **Removed metrics daemon** — The background `metrics.tick()` calls in progress indicators were removed; metric handlers are now updated inline.

### Progress Indicators

- **Elapsed time & throughput** — Both `SpinnerProgressIndicator` and `JsonProgressIndicator` now track wall-clock time and display elapsed seconds and records-per-second.

### Database Connector

- **`make_type_retriever` accepts `**kwargs`** — The abstract method, `Target`, and `NullConnector` all forward arbitrary keyword arguments to the retriever constructor, enabling database-specific options without changing the interface.
- **`make_migrator` return type fix** — Corrected the return type annotation from `TypeRetriever` to `Migrator`.
- **Node type safety in statistics** — `QueryExecutorWithStatistics` now defaults `node.type` to `"unknown"` when `None`.

### Ingestion Model

- **Sequential relationship & hook ingestion** — `DesiredIngestion.ingest_relationships` and `run_ingest_hooks` now `await` each item sequentially instead of using `asyncio.gather`, avoiding potential contention from upstream async calls filling the event loop.

### Miscellaneous

- **`NullObjectStore` no longer logs on init** — The noisy startup log line was moved to `PipelineFile` and only fires when a null store is actually in use.
- **`PipelineDefinition.initialize_for_introspection`** — Now delegates to `PipelineFile.load_pipeline_for_introspection()` for a cleaner separation of concerns.
- **`Target.connector` property** — Now delegates to `make_connector()`, which accepts overrides and logs the effective config.


nodestream/model/desired_ingestion.py
```diff
@@ -29,17 +28,12 @@ async def ingest_source_node(self, strategy: "IngestionStrategy"):
...
 
     async def ingest_relationships(self, strategy: "IngestionStrategy"):
-        await asyncio.gather(
-            *(
-                strategy.ingest_relationship(relationship)
-                for relationship in self.relationships
-            )
-        )
+        for relationship in self.relationships:
+            await strategy.ingest_relationship(relationship)
```
angelosantos4, line 35: Note when the upstream extractor has alot of asyncronous calls, this portion of code would prior yield too often to the eventloop before any relationships could even be ingested.

nodestream/databases/copy.py
```diff
@@ -20,39 +41,107 @@ def get_relationships_of_type_between(
...
+            self._relationship_counts[relationship_type] = (
+                await self.type_retriever.preview_relationship_count(relationship_type)
+            )
+
+        self.node_types.sort(key=lambda t: self._node_counts[t], reverse=True)
+        self.relationship_types.sort(
+            key=lambda t: self._relationship_counts[t], reverse=True
+        )
+
+        self._log_histogram()
```
jbristow, line 100: I don't think this is called in a test, which means none of the sorting or histogram logging ever gets tested.

I think you might want to update tests that are testing extract_records to initialize them via start.

nodestream/databases/copy.py
```diff
@@ -83,3 +170,101 @@ def convert_relationship_to_ingest(self, relationship: RelationshipWithNodes):
...
+            async def run_with_limit(coroutine: Coroutine) -> None:
+                async with semaphore:
+                    Metrics.get().increment(ACTIVE_QUERIES)
+                    try:
+                        await coroutine
+                    finally:
+                        Metrics.get().decrement(ACTIVE_QUERIES)
+
+            if coroutines:
+                await asyncio.gather(*(run_with_limit(c) for c in coroutines))
```
jbristow, line 243: Claude wrote this, and it sounds important if true.

> run_bounded wraps each coroutine with a semaphore but has no exception handling. If produce_nodes or produce_relationships raises (e.g. a database error mid-stream), asyncio.gather will propagate it out of orchestrate(), which will cause orchestrator_task to fail. The consumer loop (while True: message = await queue.get()) will then hang forever waiting for DoneObject that never arrives, because orchestrate() raised before reaching await queue.put(DoneObject). The finally: await orchestrator_task will then re-raise the original exception — but only after the consumer hangs indefinitely.

nodestream/databases/writer.py
```diff
@@ -64,3 +95,92 @@ async def finish(self, _):
...
+    async def _drain_pending_flushes(self):
+        """Wait for all in-flight background flushes to complete."""
+        if self.pending_flush_tasks:
+            await asyncio.gather(*self.pending_flush_tasks, return_exceptions=True)
+            self.pending_flush_tasks.clear()
+        self._raise_if_flush_errors()
+
+    def _raise_if_flush_errors(self):
+        if self.flush_errors:
+            raise self.flush_errors.pop(0)
```
jbristow: flush_errors is a list, and `pop(0)` has to traverse the whole list to get there.

Maybe use a collections.deque and then `self.flush_errors.popleft()`?

tests/unit/cli/commands/test_copy.py
```diff
@@ -111,5 +111,24 @@ async def test_handle_async(copy_command, mocker, basic_schema, project):
...
+    mocker.patch.object(
+        type(copy_command),
+        "has_json_logging_set",
+        new_callable=mocker.PropertyMock,
+        return_value=False,
+    )
+
     await copy_command.handle_async()
-    copy_command.run_operation.await_count == 2
+    assert copy_command.run_operation.await_count == 4
```
jbristow, line 145: At least this is asserting *something* now, but it would be better if you were testing actual results (💪 ) and not only behavior (weak).

nodestream/databases/writer.py
```diff
@@ -64,3 +95,92 @@ async def finish(self, _):
         """Close connector by calling finish method from Step"""
         await self.flush()
         await self.ingest_strategy.finish()
+
+
+class ConcurrentGraphDatabaseWriter(GraphDatabaseWriter):
```
jbristow, line 101: this isn't tested at all, you only ever test with `concurrency=1` so tests never set one of these up from scratch.

nodestream/databases/copy.py
angelosantos4: Addressed with:

```
          finally:
                # Always signal the consumer so it never hangs, even on error.
                await queue.put(DoneObject)
```

zprobst (approved):

PR #456, opened 2026-04-22
0.15.2 Fixes Bundle

angelosantos4:
This bundle rolls up four fixes and one feature into a single 0.15.2 release.

  ---

  ### fix: zombie pipeline processes on fatal error
  _`fix/zombie-pipeline-cancel-on-fatal-error`_

  When a writer step hit a fatal error, `PipelineProgressReporter.on_finish_callback` re-raised the
  exception out of the pipeline output executor. The bare `asyncio.gather()` in `Pipeline.run()`
  returned the exception but left sibling tasks — e.g. a `StreamExtractor` blocked on
  `connector.poll()` waiting for the next Kafka message — running forever. The process stayed alive
  as a zombie with nothing to do.

  **Fix:** wrap the gather in a `try/except` that cancels all tasks on any exception, waits for
  cancellation acknowledgement, then re-raises.

  **Tests added:**
  - Asserts no tasks leak after `Pipeline.run()` raises
  - Asserts the graceful channel-based drain (`StopStepExecution` → `finish()`) still completes
    before cancellation fires

  ---

  ### fix: relationship dedup key includes property names
  _`fix/relationship-dedup-key`_

  `Relationship.get_dedup_key()` used `.values()` only, so two relationships between the same nodes
  with different key property _names_ but the same _value_ (e.g. `{reason: 'foo'}` and
  `{source: 'foo'}`) hashed identically and one write was silently dropped within a batch.

  **Fix:** use `.items()` instead of `.values()`.

  Also adds `MERGE_KEY` as a `RelationshipCreationRule` option and wires `relationship_creation_rule`
  through `RelationshipInterpretation`.

  ---

  ### fix: spurious drop index ops for empty type sentinel
  _`fix/relationship-index-drift-on-empty-type`_

  `AutoChangeDetector` represents missing schema types with a sentinel `GraphObjectSchema(name="")`.
  When iterating pairs with `ignore_created_and_deleted=False`, newly created types
  (`from_type.name == ""`) were producing spurious `DropAdditionalRelationshipPropertyIndex` /
  `DropAdditionalNodePropertyIndex` operations for the empty-string type name.

  **Fix:** skip deletions when `from_type.name` is empty; skip additions when `to_type.name` is empty.

  ---

  ### feat: schema provenance + explain schema CLI
  _`add_schema_inspector`_

  Adds `SchemaProvenance` tracking to `SchemaExpansionCoordinator` — pipelines now record which
  pipeline contributed each node/relationship type to the schema via a `pipeline_context()` context
  manager. New `explain-schema` CLI command surfaces per-pipeline schema contributions. Exit codes
  added across all CLI commands for consistent error reporting.

### DIFF
--- a/nodestream/model/graph_objects.py
+++ b/nodestream/model/graph_objects.py
@@ -178,7 +178,7 @@ class Relationship(DeduplicatableObject):
         self.properties.update(other.properties)
 
     def get_dedup_key(self) -> tuple:
-        return tuple(sorted(self.key_values.values()))
+        return tuple(sorted(self.key_values.items()))
 
 
 @dataclass(slots=True)
--- a/nodestream/pipeline/pipeline.py
+++ b/nodestream/pipeline/pipeline.py
@@ -525,6 +525,17 @@ class Pipeline(ExpandsSchemaFromChildren):
         await current_output.done()
 
         # Run the pipeline by running all the steps and the pipeline output
-        # concurrently. This will block until all steps are finished.
-        # Wait for all the executors to finish.
-        await gather(*(create_task(executor.run()) for executor in executors))
+        # concurrently. If any executor raises (e.g. on_finish_callback re-raises
+        # a fatal writer error), cancel the remaining tasks so that blocking
+        # extractors (e.g. a StreamExtractor polling Kafka forever) don't keep
+        # the process alive as a zombie after the pipeline has already failed.
+        tasks = [create_task(executor.run()) for executor in executors]
+        try:
+            await gather(*tasks)
+        except Exception:
+            for task in tasks:
+                task.cancel()
+            # Wait for all tasks to acknowledge cancellation before re-raising,
+            # so the event loop is clean when the caller handles the exception.
+            await gather(*tasks, return_exceptions=True)
+            raise
--- a/nodestream/pipeline/pipeline.py
+++ b/nodestream/pipeline/pipeline.py
@@ -1,5 +1,5 @@
-from abc import ABC, abstractmethod
 import asyncio
+from abc import ABC, abstractmethod
 from asyncio import create_task, gather
 from dataclasses import dataclass, field
 from enum import Enum, auto
--- a/nodestream/pipeline/pipeline.py
+++ b/nodestream/pipeline/pipeline.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+import asyncio
 from asyncio import create_task, gather
 from dataclasses import dataclass, field
 from enum import Enum, auto
@@ -532,10 +533,24 @@ class Pipeline(ExpandsSchemaFromChildren):
         tasks = [create_task(executor.run()) for executor in executors]
         try:
             await gather(*tasks)
-        except Exception:
+        except BaseException:
             for task in tasks:
                 task.cancel()
-            # Wait for all tasks to acknowledge cancellation before re-raising,
-            # so the event loop is clean when the caller handles the exception.
-            await gather(*tasks, return_exceptions=True)
+            # Wait for all tasks to acknowledge cancellation before re-raising.
+            # Timeout prevents a misbehaving task from hanging the cleanup indefinitely.
+            # Secondary exceptions from sibling tasks are logged so they aren't silently dropped.
+            try:
+                results = await asyncio.wait_for(
+                    gather(*tasks, return_exceptions=True), timeout=30
+                )
+                for result in results:
+                    if isinstance(result, Exception):
+                        reporter.logger.warning(
+                            "Secondary exception during pipeline cancellation cleanup",
+                            exc_info=result,
+                        )
+            except TimeoutError:
+                reporter.logger.warning(
+                    "Pipeline cleanup timed out after 30s — some tasks may not have acknowledged cancellation"
+                )
             raise
--- a/nodestream/schema/migrations/auto_change_detector.py
+++ b/nodestream/schema/migrations/auto_change_detector.py
@@ -521,22 +521,36 @@ class AutoChangeDetector:
     def detect_node_index_changes(self):
         for pair in self.get_node_pairs(ignore_created_and_deleted=False):
             deleted_indexes, added_indexes = pair.get_index_drift()
-            self.deleted_node_property_indexes.update(
-                (pair.to_type.name, idx) for idx in deleted_indexes
-            )
-            self.added_node_property_indexes.update(
-                (pair.to_type.name, idx) for idx in added_indexes
-            )
+            # A missing from_type is represented by a sentinel with name="".
+            # Dropping an index from a type that didn't exist is nonsensical —
+            # skip deletions for newly created types (no prior state to drop from).
+            if pair.from_type.name:
+                self.deleted_node_property_indexes.update(
+                    (pair.to_type.name, idx) for idx in deleted_indexes
+                )
+            # A missing to_type means the type is being dropped; its indexes
+            # vanish with it, so skip additions for deleted types.
+            if pair.to_type.name:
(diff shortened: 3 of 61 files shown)


nodestream/schema/migrations/auto_change_detector.py
```diff
@@ -518,24 +518,42 @@ def detect_relationship_property_changes(self):
...
+        added_index_set: set,
+    ) -> None:
+        """Apply index drift from a type pair into the given tracking sets.
+
+        A sentinel from_type (name="") means the type is newly created — there
+        is no prior state to drop indexes from. A sentinel to_type means the
+        type is being deleted — its indexes vanish with it, so no additions.
+        """
+        deleted_indexes, added_indexes = pair.get_index_drift()
+        if pair.from_type.name:
```
AGhafaryy, line 534: This should be instead:

```
if pair.from_type.name and pair.to_type.name:
```

nodestream/schema/migrations/auto_change_detector.py
```diff
@@ -518,24 +518,42 @@ def detect_relationship_property_changes(self):
             for old, new in renamed_properties:
                 self.renamed_relationship_properties.add((pair.to_type.name, old, new))
 
-    def detect_node_index_changes(self):
-        for pair in self.get_node_pairs(ignore_created_and_deleted=False):
-            deleted_indexes, added_indexes = pair.get_index_drift()
-            self.deleted_node_property_indexes.update(
-                (pair.to_type.name, idx) for idx in deleted_indexes
+    def _apply_index_drift(
```
AGhafaryy, line 521: branch is gated only on pair.from_type.name. When a type with additional  property indexes is deleted this writes tuples like ("", "email") into  deleted_node_property_indexes, generating migration ops DropAdditionalNodePropertyIndex(node_type="", field_name="email").

nodestream/schema/migrations/auto_change_detector.py
AGhafaryy: Also can you add a test for that case i dont think i see one

nodestream/schema/state.py
```diff
@@ -817,22 +888,18 @@ def connect(
         self.unbound_adjacencies.append(unbound)
 
     """
-        For child expanders, this method is called to maintain the context layer. 
+        For child expanders, this method is called to maintain the context layer.
     """
 
     @contextmanager
-    def aquire_context(self, should_be_distinct: bool):
+    def acquire_context(self, should_be_distinct: bool):
```
AGhafaryy, line 895: do we know if other plugins are use the old nomenclature (suspecting neo4j plugin) ? this might be a breaking change

nodestream/pipeline/pipeline.py
```diff
@@ -525,6 +525,17 @@ async def run(self, reporter: PipelineProgressReporter):
...
-        # Wait for all the executors to finish.
-        await gather(*(create_task(executor.run()) for executor in executors))
+        # concurrently. If any executor raises (e.g. on_finish_callback re-raises
+        # a fatal writer error), cancel the remaining tasks so that blocking
+        # extractors (e.g. a StreamExtractor polling Kafka forever) don't keep
+        # the process alive as a zombie after the pipeline has already failed.
+        tasks = [create_task(executor.run()) for executor in executors]
+        try:
+            await gather(*tasks)
+        except Exception:
```
AGhafaryy: this doesnt catch something outside pipeline.run() telling it to stop. the exeception will arrive as CancelledError, and this block is skipped ... continuing the zombie scenario. Do you expect this to happen ?

nodestream/pipeline/pipeline.py
```diff
@@ -525,6 +525,17 @@ async def run(self, reporter: PipelineProgressReporter):
...
+            for task in tasks:
+                task.cancel()
+            # Wait for all tasks to acknowledge cancellation before re-raising,
+            # so the event loop is clean when the caller handles the exception.
+            await gather(*tasks, return_exceptions=True)
```
AGhafaryy: this gather has no timeout, so we might face the same hang ... Wrap in asyncio.wait_for(..., timeout=N) ??

nodestream/model/graph_objects.py
```diff
@@ -178,7 +178,7 @@ def update(self, other: "Relationship"):
         self.properties.update(other.properties)
 
     def get_dedup_key(self) -> tuple:
```
AGhafaryy, line 180: same question about downstream customers/plugins here using this .. ?

nodestream/schema/state.py
```diff
@@ -708,7 +738,12 @@ def bind(
         return adjacency, cardinality
 
 
-@dataclass(slots=True, frozen=True)
+# NOTE: frozen=True was intentionally removed from this dataclass.  The
+# pipeline_context() context manager must be able to mutate _current_pipeline
+# at runtime to track which pipeline is currently being expanded.  A frozen
+# dataclass would raise a FrozenInstanceError on that assignment, so mutability
+# is required here.
+@dataclass(slots=True)
```
AGhafaryy, line 746: why is frozen=true dropped ? How about moving _current_pipeline to module level ? 

```
 _current_pipeline: ContextVar[Optional[str]] = ContextVar("...", default=None).
  pipeline_context() sets/resets via .set() / .reset(token). 
```

Restore frozen=True on the dataclass. 

opinions ?

nodestream/schema/state.py
```diff
@@ -724,11 +759,22 @@ class SchemaExpansionCoordinator:
...
     )
+    provenance: SchemaProvenance = field(default_factory=SchemaProvenance)
+    _current_pipeline: Optional[str] = None
     # Maps node types to their additional types for relationship expansion
     additional_types_map: LayeredDict[str, Tuple[str, ...]] = field(
         default_factory=LayeredDict
     )
 
+    @contextmanager
+    def pipeline_context(self, pipeline_name: Optional[str]):
```
AGhafaryy, line 770: this context manager might not be concurrency safe

nodestream/pipeline/pipeline.py
AGhafaryy: here only the first exception is shown and the rest is dropped.

nodestream/schema/state.py
angelosantos4: acquire_context is only called from within state.py itself (ExpandsSchema.expand_schema at line 1071). No plugin calls it directly. This is an entrypoint for the schema expansion.

nodestream/schema/state.py
angelosantos4: The mutation is intentional — _current_pipeline needs to be set per-pipeline as the coordinator walks through all pipelines sequentially, so provenance recording (lines 810/834/852) knows which pipeline to attribute each schema object to. There's no way around the mutation without restructuring provenance tracking entirely. The ContextVar approach is clean but solves a concurrency problem that doesn't exist here: a single coordinator instance is created and walked sequentially, never shared across concurrent tasks. Restoring frozen=True would just move the FrozenInstanceError from the dataclass to runtime. The note in the comment above the dataclass explains the reasoning.

nodestream/schema/state.py
angelosantos4: Schema expansion is fully sequential — one coordinator is created per expand_schema_for_scopes call and pipelines are expanded in a for-loop, never with gather or create_task. So pipeline_context mutating _current_pipeline on that instance carries no concurrency risk in practice.

nodestream/model/graph_objects.py
angelosantos4: This was unchanged, also this is in the core framework. No names have been modified in this section, this is a change to the deduplication method.

AGhafaryy (approved):

PR #457, opened 2026-05-27
0.16 - Copy Command Rework

angelosantos4:
## Summary

Refactors the `nodestream copy` pipeline to support plugin-driven concurrent copying with histogram-guided sharding and a clean retriever contract.

### What changed

**New `TypeRetriever` ABC** (`nodestream/databases/copy.py`)

Introduces a `TypeRetriever` abstract base class that plugins implement to own all type decomposition, shard splitting, histogram computation, and concurrency strategy. The `Copier` only sees two async generators — `fetch_nodes(schema)` and `fetch_relationships(schema)` — and drives them through a producer/consumer queue when `concurrency_limit > 1`. Plugin internals are completely invisible to core.

**Histogram-driven sharding**

At startup, `Copier` calls `build_histogram(schema)` to get record counts per type. When `shard_size` is set, each type is split into fixed-size shards that run as independent concurrent coroutines. This means:
- All 40 concurrency slots stay busy — when a large type's last shard finishes, the next type's shards fill in immediately
- No single large type can starve smaller ones
- Types without a schema key field fall back to `elementId`-based sharding automatically

**Flat retriever contract** (`nodestream/cli/commands/copy.py`, `nodestream/cli/operations/run_copy.py`)

CLI-level flags (`--retriever-option key=value`) are merged into a single flat dict passed to `make_type_retriever`. Core no longer forwards plugin-specific kwargs explicitly — each plugin pops what it understands and ignores the rest. Neptune and future plugins are unaffected.

**`expand_schema_for_copy` error visibility** (`nodestream/project/pipeline_definition.py`)

Replaced a bare `except Exception: pass` with `logger.warning(..., exc_info=True)` so silent schema expansion failures are now visible in logs.

### New CLI options

| Flag | Description |
|---|---|
| `--retriever-option key=value` | Pass arbitrary key/value options to the plugin's `TypeRetriever` (repeatable) |
| `--flush-concurrency N` | Number of parallel write lanes in the writer |
| `--step-outbox-size N` | Buffer size between pipeline steps |

### Performance

Benchmarked on a production Neo4j graph (~135M relationships, `sample_ratio=50`, `latest_hours=24`):

| | Before | After |
|---|---|---|
| Avg throughput | ~400 r/s | ~1,000–1,200 r/s |
| Histogram build | ~22 min (sequential COUNTs) | ~1 min (`asyncio.gather`) |
| Concurrency utilisation | Low (single coroutine per type) | High (40 shards in flight) |

### Test plan

- [ ] `poetry run pytest tests/unit/databases/test_copy.py` — Copier, TypeHistogram, concurrent path
- [ ] `poetry run pytest tests/unit/cli/operations/test_run_copy.py` — CLI wiring
- [ ] `poetry run pytest tests/unit/pipeline/test_pipeline_file.py` — schema-only load path
- [ ] `poetry run ruff check . && poetry run black --check . && poetry run isort --check-only .` — lint clean

### DIFF
--- a/nodestream/cli/commands/copy.py
+++ b/nodestream/cli/commands/copy.py
@@ -188,19 +188,23 @@ class Copy(NodestreamCommand):
         node_filter = []
         rel_filter = []
 
-        all_node_names = {n.name for n in schema.nodes}
-        for n in self.option("node") or []:
-            if n in all_node_names:
-                node_filter.append(n)
+        all_node_names = {nodeType.name for nodeType in schema.nodes}
+        for nodeTypeName in self.option("node") or []:
+            if nodeTypeName in all_node_names:
+                node_filter.append(nodeTypeName)
             else:
-                logger.warning("Unknown node type %r — skipping", n)
+                logger.warning("Unknown node type %r — skipping", nodeTypeName)
 
-        all_rel_names = {r.name for r in schema.relationships}
-        for r in self.option("relationship") or []:
-            if r in all_rel_names:
-                rel_filter.append(r)
+        all_rel_names = {
+            relationshipType.name for relationshipType in schema.relationships
+        }
+        for relationshipTypeName in self.option("relationship") or []:
+            if relationshipTypeName in all_rel_names:
+                rel_filter.append(relationshipTypeName)
             else:
-                logger.warning("Unknown relationship type %r — skipping", r)
+                logger.warning(
+                    "Unknown relationship type %r — skipping", relationshipTypeName
+                )
 
         return schema.filtered(node_filter=node_filter, relationship_filter=rel_filter)
 
--- a/nodestream/cli/commands/copy.py
+++ b/nodestream/cli/commands/copy.py
@@ -104,10 +104,10 @@ class Copy(NodestreamCommand):
             flag=False,
         ),
         option(
-            "node-only",
+            "preload-nodes",
             description=(
-                "Copy only nodes — skip relationship fetching entirely. "
-                "Useful when you want to seed nodes without their adjacencies."
+                "Copy nodes first, then relationships. "
+                "Useful when the destination needs nodes seeded before adjacencies are written."
             ),
             flag=True,
         ),
@@ -123,7 +123,7 @@ class Copy(NodestreamCommand):
             try:
                 from_target = self.get_taget_from_user(project, "from")
                 to_target = self.get_taget_from_user(project, "to")
-                node_only = bool(self.option("node-only"))
+                preloadNodes = bool(self.option("preload-nodes"))
 
                 full_schema = project.make_schema_for_copy(
                     include_additional_types=False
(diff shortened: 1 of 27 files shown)


AGhafaryy:
Review — correctness, reliability & performance
  
  Reviewed against the PR diff. Flagging real issues only (P0–P2); naming/style nits omitted.

  P1 — RelationshipWithNodes.into_ingest() changes ingest semantics for all callers, not just copy

  nodestream/model/graph_objects.py

Instead of this:
  -            node_creation_rule=NodeCreationRule.MATCH_ONLY,
  -            relationship_creation_rule=RelationshipCreationRule.EAGER,
  
  Do this:
  +            node_creation_rule=self.to_side_node_creation_rule,
  +            relationship_creation_rule=self.relationship_creation_rule,

  The to-side node rule was hardcoded to MATCH_ONLY; it now reads self.to_side_node_creation_rule, which defaults to EAGER. into_ingest() is a shared-model method used by the normal ingest pipeline, not
  only by copy. Any RelationshipWithNodes built without explicitly setting the rule now creates endpoint nodes (with key-only data) instead of matching existing ones — risking skeletal/duplicate
  destination nodes. Please either keep the model default at MATCH_ONLY and have the copy extractors set EAGER explicitly, or confirm + test this as an intended global change. No test asserts
  source_node_creation_rule wiring either.

  P1 — expand_schema_for_copy silently swallows per-pipeline failures → silent partial copy

  nodestream/project/pipeline_definition.py
  
  
Add this:
  +        except Exception:
  +            logger.warning("Schema collection failed for pipeline '%s'; skipping.", self.name, exc_info=True)

  If one pipeline fails schema collection (import error, env/cred touch), its node/relationship types are silently dropped from the schema make_schema_for_copy returns. The copy then quietly skips that
  data and exits 0 — data loss that looks like success, the worst failure mode for a copy tool. Suggest an aggregate "N of M pipelines failed; these types excluded" summary, and/or fail by default with an
  explicit --allow-partial-schema opt-out.

  P2 — Copier.extract_records creates one task per shard up front (unbounded)

  nodestream/databases/copy.py

  async for extractor in self.type_retriever.fetch_extractors():
      tasks.append(asyncio.create_task(run_extractor(extractor)))
  await asyncio.gather(*tasks)

  The semaphore bounds execution, not task creation. With sharding, a large source can yield tens of thousands of Task+Extractor objects all created before gather. Prefer a fixed pool of concurrency_limit
  workers pulling from a bounded queue so live tasks are capped by concurrency, not shard count.

  P2 — bounded record_queue + early consumer exit can leak/block producers

  nodestream/databases/copy.py

  record_queue = asyncio.Queue(maxsize=self.queue_size) (wired to --step-outbox-size). If the downstream closes and the generator is abandoned without aclose(), run_extractor tasks can block forever on
  await record_queue.put(...) and await producer_task never resolves. Cancel the producer in the finally before awaiting it (suppressing CancelledError).

  P2 — relationship flush changed sequential → concurrent; retry-storm risk under overlap

  nodestream/databases/debounced_ingest_strategy.py


Add this: 
  +    def flush_relationship_updates(self):
  +        # Run all relationship-shape flushes concurrently. Neo4j may deadlock ... retry loop handles them.
  +        return asyncio.gather(*update_coroutines)

  The old code serialized relationship-shape flushes specifically to avoid Neo4j deadlocks. Running them concurrently and relying on deadlock-retry can be a net loss under heavy endpoint overlap (many
  flushes deadlock + retry together). Please confirm the executor's retry path has backoff, and ideally benchmark the overlap-heavy case.
  
  
  Compound Engineering PR Review. Feel free to push back on things that don't make sense

AGhafaryy (approved):

PR #458, opened 2026-06-09
fix: zombie pipeline cleanup, !pluck YAML tag support, metrics report…

angelosantos4:
## Summary

- **Zombie pipeline fix**: Catch `BaseException` (not just `Exception`) in `ProcessRecordsState` and `EmitOutstandingRecordsState` so `CancelledError` transitions through `StopStepExecution` and always calls `step.finish()` — preventing Kafka consumers / IDPS credential-refresh threads from leaking after fatal pipeline failure
- **Metrics Report / Pipeline Completed fix**: Exclude `pipeline_output` executor (index 0) from task cancellation so `on_finish_callback` always fires, restoring the end-of-pipeline metrics report
- **`!pluck` YAML tag fix**: `wrap_unloaded_tag` in `file_io.py` now handles mapping and sequence YAML nodes (previously only scalar), fixing `ConstructorError: expected a scalar node, but found mapping` when running `nodestream explain schema` on pipelines that use `!pluck` with dict arguments
- **Version bump**: `0.15.2` → `0.15.3`

## Root Cause (Zombie Pipeline)

`CancelledError` is a `BaseException`, not an `Exception`. When `pipeline.run()` cancels a `StreamExtractor` task mid-flight, `CancelledError` is raised inside `ProcessRecordsState` / `EmitOutstandingRecordsState`. Those states caught `except Exception`, so `CancelledError` escaped — short-circuiting the state machine before `StopStepExecution` where `step.finish()` is called. Any resource the extractor holds (Kafka consumer, IDPS thread) remained alive after "Pipeline Completed".

```
StartStepState
    → ProcessRecordsState       ← CancelledError escapes here (was: except Exception)
    → EmitOutstandingRecordsState  ← or here
    → StopStepExecution         ← step.finish() lives here — never reached on 0.15.2
```

## Changes

| File | Change |
|---|---|
| `nodestream/pipeline/pipeline.py` | Catch `BaseException` in state machine; exclude `pipeline_output` task from cancellation |
| `nodestream/pipeline/step.py` | Add `cancellation_cause: Optional[BaseException]` to `StepContext` |
| `nodestream/file_io.py` | Handle mapping/sequence nodes in `wrap_unloaded_tag` |
| `pyproject.toml` | Version `0.15.2` → `0.15.3` |

## Test Plan

- [ ] `pytest tests/unit/pipeline/test_pipeline.py` — `test_pipeline_fatal_error_drains_completed_steps_before_cancellation` is the regression gate for the zombie fix
- [ ] `pytest tests/unit/` — full unit suite green
- [ ] Manual: run a pipeline with a fatal writer error and confirm `finish()` is called on the extractor step and metrics report appears at end of run

### DIFF
--- a/nodestream/pipeline/pipeline.py
+++ b/nodestream/pipeline/pipeline.py
@@ -344,8 +344,11 @@ class StopStepExecution(StepExecutionState):
         except Exception as e:
             self.context.report_error("Error stopping step", e)
 
-        # We are done regarless of what happens. There is no next state.
-        Metrics.get().decrement(STEPS_RUNNING)
+        finally:
+            # Decrement in finally so the gauge is always accurate even if
+            # finish()/output.done() raises CancelledError or another
+            # BaseException that bypasses the except clause above.
+            Metrics.get().decrement(STEPS_RUNNING)
 
         # If a prior state caught a CancelledError (or other BaseException) to
         # allow finish() to run, re-raise it now so the executor task is
@@ -559,10 +562,6 @@ class Pipeline(ExpandsSchemaFromChildren):
         # raises (e.g. on_finish_callback re-raises a fatal writer error), cancel
         # the step tasks so that blocking extractors (e.g. a StreamExtractor
         # polling Kafka forever) don't keep the process alive as a zombie.
-        # pipeline_output_task is intentionally excluded from cancellation — it
-        # must finish so that on_finish_callback (Metrics Report, "Pipeline
-        # Completed") always runs. Its input channel is already closed by the
-        # time any step raises, so it drains and exits on its own.
         try:
             # StopStepExecution calls step.finish() before output.done(), so by
             # the time pipeline_output receives DoneObject and calls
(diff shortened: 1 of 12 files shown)


nodestream/pipeline/pipeline.py
```diff
@@ -311,6 +324,12 @@ class StopStepExecution(StepExecutionState):
 
     async def execute_until_state_change(self) -> Optional[StepExecutionState]:
         try:
+            # Run finish() before signaling downstream. This ensures that any
+            # metrics accumulated during finish() (e.g. NODES_UPSERTED from a
+            # final flush) are recorded before pipeline_output calls
+            # on_finish_callback to emit the Metrics Report.
+            await self.step.finish(self.context)
+
```
AGhafaryy, line 332: The try still catches only Exception. If CancelledError (a BaseException) is injected during finish()/output.done(), it bypasses the except and
    propagates before decrement(STEPS_RUNNING), leaving the gauge incremented. Across repeated failures this monotonically inflates the very metric used to detect stuck steps. maybe decrement in a finally.

nodestream/pipeline/pipeline.py
```diff
@@ -482,25 +512,23 @@ async def run(self, reporter: PipelineProgressReporter):
...
-        # scheduled is the one that calls the on_start_callback.
+        # pipeline_output_task is created first so that on_start_callback fires
+        # before any step coroutine gets a chance to run. asyncio schedules tasks
+        # in creation order within a single event-loop tick, so creating this task
+        # before the step tasks guarantees the ordering.
         current_input, current_output = channel(
             self.step_outbox_size, current_output_name, current_input_name
         )
-        executors.append(Executor.pipeline_output(current_input, reporter))
+        pipeline_output_task = create_task(
```
AGhafaryy, line 528: pipeline_output_task is excluded from cancellation AND from the 30s timeout — the outer gather(pipeline_output_task, *step_tasks) is unbounded. PipelineOutputStopState calls on_finish_callback
    deliberately without swallowing exceptions; if that user callback blocks (slow/hung metrics sink — exactly the "final-flush metrics" this PR adds), or if a step is cancelled mid-finish() before
    reaching output.done() so no DoneObject is ever emitted, the output task blocks on input.get() forever. This relocates the zombie hang the PR targets from the Kafka extractor to the output task.

AGhafaryy (approved):

